PR #10, opened 2019-08-06
Feature/41 programme hero

nicklee:


rca/static_src/sass/base/_typography.scss
```diff
@@ -34,10 +34,6 @@ blockquote {
     text-transform: inherit;
 }
 
-h1 {
-    @include font-size(xxl);
-}
-
```
chris-lawton: we don't need this?

nicklee: We're going to move these classes into a rich-text component, so that we're not having to override sizes on every h tag with heading heading--modifier.

PR #46, opened 2019-08-13
z-index refactor to allow for images to sit in front of grid lines

nicklee:


rca/static_src/sass/components/_logo.scss
```diff
@@ -25,6 +25,7 @@
     }
 
     .header-content-dark .header & {
+        // sass-lint:disable-line force-element-nesting
```
chris-lawton: this can go if it's disabled in `.sass-lint.yml`?

PR #50, opened 2019-08-13
Feature/58 related content fe

chris-lawton:
@kevinhowbrook - i've removed bits from the `relatedcontent.html` template as I don't think we'll be needing it. Can you just take a look over and make sure there's nothing we need that i've removed?


kevinhowbrook:
@chris-lawton yea looks good to me, I've yet to wire up the related programmes section yet but looks like we have what we need for it

PR #53, opened 2019-08-13
Feature/59 carousel fe

nicklee:


nicklee:
Frank said that the course should no longer appear, as a lot of them will have been on a course that is no longer active.

PR #57, opened 2019-08-14
Feature/66 curriculum intro

chris-lawton:


rca/static_src/sass/components/_image-caption-link.scss
```diff
@@ -0,0 +1,97 @@
...
+
+    &__caption-icon {
+        margin-right: ($gutter / 2);
+    }
+
+    &__image-icon {
+        position: absolute;
+        left: 50%;
+        top: 50%;
+        transform: translate(-50%, -50%);
```
nicklee: any reason not to use translate3d to ensure the graphics card is doing the work instead of CPU? (better performance)

PR #61, opened 2019-08-15
Fees and funding initial setup

nicklee:


rca/static_src/sass/components/_fees.scss
```diff
@@ -0,0 +1,76 @@
...
+                    padding: $gutter 0 ($gutter * 1.5);
+                }
+            }
+
+            #{$root}__item--title {
+                grid-column: 1 / span 1;
+                padding: $gutter 0;
+
+                @include media-query(large) {
+                    grid-column: 1 / span 1;
```
chris-lawton: think this can go and just fall back to the first `grid-column: 1 / span 1`

nicklee: fixed, thanks

PR #71, opened 2019-08-16
Feature/72 requirements be

kevinhowbrook:
Adds field for the review section of the programme page


rca/project_styleguide/templates/patterns/organisms/programme_tabs/requirements.html
```diff
@@ -1,5 +1,8 @@
 <section class="section section--start bg bg--dark">
     Requirements content here
+{{ page.requirements_subtitle }}
+{{ page.requirements_text }}
+{{ page.requirements_blocks }}
```
tm-kn: Even if it's sample, can you please use `{% include_block %}` in case front-end gets to this?

rca/project_styleguide/templates/patterns/organisms/programme_tabs/requirements.html
```diff
@@ -1,5 +1,8 @@
 <section class="section section--start bg bg--dark">
     Requirements content here
+{{ page.requirements_subtitle }}
```
tm-kn: Can you sorround variables with if-blocks if values are not mandatory?

PR #88, opened 2019-08-20
Feature/73 fees funding be

kevinhowbrook:
Note: I've changed some of the fees fields here when compared to the field spec. After looking at the admin UI the field spec wasn't the nicest solution.


rca/project_styleguide/templates/patterns/organisms/programme_tabs/fees-and-funding.html
```diff
@@ -1,87 +1,110 @@
 <section class="section section--start section--end bg bg--dark">
     <header class="section__header grid">
-        <h2 class="section__heading section__heading--primary heading heading--two">{{ tab_section.title }}</h2>
+        <h2 class="section__heading section__heading--primary heading heading--two">Fees & funding</h2>
```
tm-kn: `&amp;`? I dont think Django converts unless it's a variable.

rca/utils/blocks.py
```diff
@@ -7,6 +7,17 @@
 from wagtail.snippets.blocks import SnippetChooserBlock
 
 
+class FeeBlock(blocks.StructBlock):
+    location = blocks.CharBlock(max_length=120, help_text="e.g. Home and EU")
+    subsidised = blocks.BooleanBlock(required=False)
+    per_year_cost = blocks.CharBlock(
+        max_length=120, help_text="e.g. £14,200 per year", required=False
+    )
+    total_cost = blocks.CharBlock(
```
tm-kn: I feel weird about not using number fields for this. You can achieve the number formatting with https://docs.djangoproject.com/en/2.2/ref/contrib/humanize/#intcomma . It will result in editors filling cost in different formats.

rca/utils/blocks.py
```diff
@@ -7,6 +7,17 @@
 from wagtail.snippets.blocks import SnippetChooserBlock
 
 
+class FeeBlock(blocks.StructBlock):
+    location = blocks.CharBlock(max_length=120, help_text="e.g. Home and EU")
```
tm-kn: I'd imagine this can only have two values. This should go back to RCA. I cannot imagine that editors will be disciplined enough to stick to the same values.  We should just use a select box or a separate taxonomy model.

rca/project_styleguide/templates/patterns/organisms/programme_tabs/fees-and-funding.html
```diff
@@ -1,87 +1,110 @@
...
+                <div class="overview__item">
+                    {% for block in page.scholarship_information_blocks %}
+                        {% if block.value.title %}
+                            <h4 class="overview__sub-heading heading heading--five">
+                                {{ block.value.title }}
+                            </h4>
+                        {% endif %}
+                        {% if block.value.text %}
+                            <p class="overview__introduction introduction">
+                                {{ block.value.text|safe }}
```
tm-kn: Why do we need `|safe`? Can we use `|richtext`?

rca/project_styleguide/templates/patterns/organisms/programme_tabs/fees-and-funding.html
```diff
@@ -1,87 +1,110 @@
...
+                            </p>
+                        {% endif %}
+                        {% if block.value.link %}
```
tm-kn: `url` and `title` can be empty as well, right? Should check for that.

rca/project_styleguide/templates/patterns/organisms/programme_tabs/fees-and-funding.html
```diff
@@ -1,87 +1,110 @@
...
+                <div class="overview__item">
+                        {% for block in page.more_information_blocks %}
+                            {% if block.value.title %}
+                                <h4 class="overview__sub-heading heading heading--five">
+                                    {{ block.value.title }}
+                                </h4>
+                            {% endif %}
+                            {% if block.value.text %}
+                                <p class="overview__introduction introduction">
+                                    {{ block.value.text|safe }}
```
tm-kn: again, `|richtext`.

rca/utils/blocks.py
kevinhowbrook: I think in the future these may switch to be snippet based, but at the moment they are text fields because in our designs we wave 'overseas', 'overseas students', 'channel islands', etc... so for time sake I'm going to leave this as is

rca/utils/blocks.py
kevinhowbrook: I share your concern, the reason I have built this with text fields as so the values can be per month, per year, per term. Again I think it will work better using number fields and snippets for the  'per year/month' values but this won't be in this phase

PR #96, opened 2019-08-21
Feature/99 programme staff

kevinhowbrook:
I'll need to re-gen the migration for this when ready to merge


rca/project_styleguide/templates/patterns/organisms/staff-overview/staff-overview--single.html
```diff
@@ -1,8 +1,10 @@
 <div class="section__row grid staff-overview staff-overview--single">
     <h2 class="heading heading--four staff-overview__heading">Staff</h2>
-    {% include "patterns/molecules/staff-detail/staff-detail.html" %}
-    <a class="staff-overview__view-all link link--primary link--link body body--two" href="#">
-        <span class="link__label">See all programme staff</span>
+    {% for item in staff %}
```
tm-kn: We probably should pass a queryset from Python that has `select_related('image')`.

rca/programmes/models.py
```diff
@@ -160,6 +185,12 @@ class ProgrammePage(BasePage):
         [("slide", GalleryBlock())], blank=True, verbose_name="Programme gallery"
     )
 
+    # Staff
+    staff_link = models.URLField(blank=True)
+    staff_link_text = models.CharField(
+        max_length=125, blank=True, help_text="E.g. 'See all programe staff'"
```
tm-kn: programe -> programme

rca/programmes/models.py
```diff
@@ -64,6 +64,31 @@ class ProgrammePageCareerOpportunities(Orderable):
...
+    role = models.CharField(max_length=125, blank=True)
+    description = models.TextField(blank=True)
+    link = models.URLField(blank=True)
+    link_text = models.CharField(max_length=125, blank=True)
+    panels = [
+        ImageChooserPanel("image"),
+        FieldPanel("name"),
+        FieldPanel("role"),
+        FieldPanel("description"),
+        FieldPanel("link"),
```
tm-kn: can we define `__str__` please?

rca/project_styleguide/templates/patterns/organisms/programme_tabs/programme_overview.html
```diff
@@ -53,10 +53,15 @@ <h3 class="section__heading section__heading--secondary heading heading--four">
 </section>
 
 <section class="section bg bg--light">
-
-    {% include "patterns/organisms/staff-overview/staff-overview.html" %}
-
-    {% include "patterns/organisms/staff-overview/staff-overview--single.html" %}
+    {% with page.related_staff.all as staff %}
```
tm-kn: Should we prefer the more modern format? https://docs.djangoproject.com/en/2.2/ref/templates/builtins/#with

`{% with staff=page.related_staff.all %}`

rca/programmes/models.py
```diff
@@ -410,6 +448,10 @@ def clean(self):
             errors["curriculum_video"].append(
```
tm-kn: This should be more reliable.

```
from wagtail.embeds import embeds
from wagtail.embeds.exceptions import EmbedException

try:
    embed = embeds.get_embed(self.curriculum_video)
except EmbedException:
    errors["curriculum_video"].append('invalid embed URL')
else:
    # not sure if this metadata is saved.
    if embed.provider_name.lower() !== 'youtube':
        # Only YouTube videos are supported for this field
```

https://github.com/wagtail/wagtail/blob/d6e4072e26ab7291d2aa56c4e87c0d82c7fb9030/wagtail/embeds/models.py#L29

rca/project_styleguide/templates/patterns/organisms/staff-overview/staff-overview--single.html
kevinhowbrook: Thanks @tm-kn, I've added this [here](https://github.com/torchbox/rca-wagtail-2019/pull/96/files#diff-7e356f728d6d9ada8c4d7be20f54d5fcR573) if you could confirm it

PR #100, opened 2019-08-21
Feature/92 apply be

kevinhowbrook:
Adds fields for the 'Apply' section of the programme page
(will need migrations tweaking depending on what is merged before hand)


rca/project_styleguide/templates/patterns/atoms/image/image.html
```diff
@@ -1,7 +1,7 @@
 {% load wagtailimages_tags %}
 
-{% image item.image fill-100x37 as imageSmall %}
-{% image item.image fill-1134x595 as imageLarge %}
+{% image image fill-100x37 as imageSmall %}
```
tm-kn: Can we not use camel case for variable names please?

kevinhowbrook: I'll alert FE of this.

PR #103, opened 2019-08-21
Add hero and strapline cta fields

kevinhowbrook:


rca/home/models.py
```diff
@@ -12,18 +15,71 @@ class HomePage(BasePage):
...
+        get_image_model_string(),
         null=True,
+        blank=True,
         on_delete=models.SET_NULL,
         related_name="+",
     )
+    hero_colour_option = models.CharField(
+        max_length=1,
+        choices=(
+            ("1", "Light text on a dark image"),
```
tm-kn: If we are going to use numbers, why not just use `PositiveSmallIntegerField`?

kevinhowbrook:
Closing in favour of https://github.com/torchbox/rca-wagtail-2019/pull/119

rca/home/models.py
kevinhowbrook: I attempted this but it resulted in `django.db.utils.DataError: invalid input syntax for integer: ""`

tm-kn: Did you change `"1"` to `1`?

PR #131, opened 2019-08-27
Configure CI for production

kevinhowbrook:


.circleci/config.yml
```diff
@@ -82,3 +92,9 @@ workflows:
           filters:
             branches:
               only: staging
+      - deploy_production:
```
tm-kn: Should we have manual approval for production?

kevinhowbrook: Good idea.

PR #141, opened 2019-08-28
Feature/api content 100 98

kevinhowbrook:
~Note: The TODOs in this PR will be completed when https://github.com/torchbox/verdant-rca/pull/186 is merged and the api is updated to support what we need to query.~

For a more detailed description of what this PR provides, see the [implementation spec](https://github.com/torchbox/rca-wagtail-2019/blob/master/docs/implementation-specs/content-relationships.md)


rca/api_content/content.py
```diff
@@ -0,0 +1,95 @@
...
+        data = resp.json()
+
+        if data["meta"]["total_count"] > 0:
+            news_items_to_get = 2
+            events_data = parse_items_to_list(data, "Event")
+        else:
+            news_items_to_get = 3
+
+        # News
+        url = f"https://rca.ac.uk/api/v2/pages/?limit={news_items_to_get}&order=-date&type=rca.NewsItem"
```
tm-kn:
```python
from django.http.request import QueryDict

query = QueryDict(mutable=True)
query.update({
    'limit': news_items_to_get,
    'order': '-date',
    'type': 'rca.NewsItem',
})
url = f"https://rca.ac.uk/api/v2/pages/?{query.urlencode}"
```
We could employ QueryDict or [urllib](https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlparse) to avoid mistakes.

rca/api_content/content.py
```diff
@@ -0,0 +1,95 @@
...
+        resp = requests.get(url=url)
```
tm-kn: `resp` -> `response`?

rca/api_content/content.py
```diff
@@ -0,0 +1,95 @@
...
+        data = resp.json()
+        news_data = parse_items_to_list(data, "News")
```
tm-kn: I'd use `resp.raise_for_status()` to fail on error or at least report it to Sentry somehow (maybe with `logger.exception()`?).

rca/programmes/migrations/0016_add_legacy_slug.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [("programmes", "0015_remove_unwanted_fields")]
+
+    operations = [
+        migrations.AddField(
+            model_name="programmetype",
+            name="legacy_slug",
+            field=models.CharField(
+                blank=True,
+                help_text="This field should match         the value added as a slug in the current live RCA sites 'Programme' taxonomy, it's used for             relating content like news and events.",
```
tm-kn: SPACEEEEEEEEEEEEEEEES

rca/programmes/models.py
```diff
@@ -47,6 +48,13 @@ def __str__(self):
 class ProgrammeType(models.Model):
     display_name = models.CharField(max_length=128)
     slug = models.CharField(max_length=128)
+    legacy_slug = models.CharField(
+        max_length=128,
+        blank=True,
+        help_text="This field should match \
```
tm-kn: Spaces

rca/programmes/models.py
```diff
@@ -515,6 +480,28 @@ class ProgrammePage(BasePage):
...
+    def get_alumni_stories(self):
+        # TODO, once we can query by programme slug, we'll need to
+        # add the slug as prefix to the cache key
+        if not cache.get("programme_latest_alumni_stories"):
+            cache.set(
+                "programme_latest_alumni_stories",
+                content.pull_alumni_stories(),
+                settings.API_CONTENT_CACHE,
+            )
+        return cache.get("programme_latest_alumni_stories")
```
tm-kn: Won't that hit the cache again? Can you set it in the variable before the if-statement?

rca/programmes/models.py
```diff
@@ -515,6 +480,28 @@ class ProgrammePage(BasePage):
...
     search_fields = BasePage.search_fields + []
 
+    def get_alumni_stories(self):
+        # TODO, once we can query by programme slug, we'll need to
+        # add the slug as prefix to the cache key
+        if not cache.get("programme_latest_alumni_stories"):
+            cache.set(
+                "programme_latest_alumni_stories",
+                content.pull_alumni_stories(),
+                settings.API_CONTENT_CACHE,
```
tm-kn: `API_CONTENT_CACHE` is this supposed to be the timeout? Could you name this setting something more intuitive please?

rca/project_styleguide/templates/patterns/molecules/card/card_api_content.html
```diff
@@ -0,0 +1,22 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+{# This template is for content pulled from the rca api #}
+<div class="card {% if item.image %}card--image{% endif %} {% if item.person %}card--person{% endif %}">
+    <div class="card__container">
+        {% if item.link %}<a href="{{ item.link }}" class="card__image-container" aria-label="{{ item.title }}">{% endif %}
+            {% if item.image and counter != 2 %}
+                <img src="{{ item.image }}">
```
tm-kn: `alt`?

rca/project_styleguide/templates/patterns/organisms/news/news_api_content.html
```diff
@@ -0,0 +1,27 @@
...
+            <div class="news__sub-features">
+        {% endif %}
+
+        {% include "patterns/molecules/card/card_api_content.html" with counter=forloop.counter %}
+
+        {% if forloop.first %}
+            </div>
+        {% endif %}
+
+        {% if forloop.last and not forloop.first %}
```
tm-kn: shouldn't `foorloop.last` be enough? Can `forloop.first` and `forloop.last` be both true at the same time?

kevinhowbrook:
> Going in a good direction. Are you going to be call these functions via Django management commands?

I wasn't planning on running the functions via management commands... that's what the cache timeout is for, so after the timeout of 1 day, the content is pulled from the api and placed in the cache.

rca/programmes/models.py
kevinhowbrook: Isn't that `cache.get_or_set()` ?  This is just retrieving the cache value isn't it?

rca/api_content/content.py
kevinhowbrook: Neat, thanks!

rca/programmes/models.py
```diff
@@ -515,6 +483,34 @@ class ProgrammePage(BasePage):
...
+                settings.API_CONTENT_CACHE_TIMEOUT,
+            )
+        return cache.get(
+            f"{programme_type_legacy_slug}_programme_latest_alumni_stories"
+        )
+
+    def get_news_and_events(self, programme_type_legacy_slug):
+        # Use the slug as prefix to the cache key
+        if not cache.get(
+            f"{programme_type_legacy_slug}_programme_latest_news_and_events"
```
tm-kn: Can we not repeat ourselves with the same cache key? Just set a variable ;)

rca/api_content/content.py
kevinhowbrook: @tm-kn I've added this based on out slack convo but not 100% sure about it

kevinhowbrook: E.g, https://github.com/torchbox/rca-wagtail-2019/pull/141/files#diff-3a8951d04a021439a3fbc9866c9a2d16R25

rca/api_content/content.py
```diff
@@ -0,0 +1,132 @@
...
+    query = query.urlencode()
+    news_url = f"https://rca.ac.uk/api/v2/pages/?{query}"
+
+    try:
+        response = requests.get(url=news_url)
+        response.raise_for_status()
+    except requests.exceptions.HTTPError:
+        logger.exception("Error occured when fetching news data")
+
+    data = response.json()
```
tm-kn: I do not think you want to call this after you had a HTTP error. You want to use `else` block or put it in the `try` block.

rca/home/models.py
```diff
@@ -174,6 +177,24 @@ def clean(self):
...
+                cache_key, news_and_events_data, settings.API_CONTENT_CACHE_TIMEOUT
+            )
+        return news_and_events_data
+
+    def get_alumni_stories(self):
+        cache_key = "latest_alumni_stories"
+        stories_data = cache.get(cache_key)
+        if not stories_data:
+            stories_data = content.pull_alumni_stories()
+            cache.set(cache_key, stories_data, settings.API_CONTENT_CACHE_TIMEOUT)
```
tm-kn: You should not set cache if code inside `pull_alumni_stories` failed. If I were you, I'd raise a custom exception in `pull_alumni_stories` or check `if stories_data is None` (and ensure that is returned from the method when failure occurs).

rca/api_content/content.py
kevinhowbrook: I've improved the exception handling so this is ready for re-review, thanks.

rca/home/models.py
kevinhowbrook: I've improved the exception handling so this is ready for re-review, thanks.

rca/api_content/content.py
```diff
@@ -0,0 +1,148 @@
...
+    query = query.urlencode()
+    url = f"https://rca.ac.uk/api/v2/pages/?{query}"
+
+    try:
+        response = requests.get(url=url)
+        response.raise_for_status()
+        logger.info("pulling Alumni Stories from API")
+    except requests.exceptions.HTTPError:
+        error_text = "Error occured when fetching alumni stories data"
+        logger.info(error_text)
```
tm-kn: Why are we not using `logger.exception`?

rca/home/models.py
```diff
@@ -174,6 +177,32 @@ def clean(self):
         if errors:
             raise ValidationError(errors)
 
+    def get_news_and_events(self):
+        cache_key = "latest_news_and_events"
+        news_and_events_data = cache.get(cache_key)
+        if not news_and_events_data:
```
tm-kn: What if this is an empty list? We need to check whether the data is in cache, not whether the value returned is falsy, right?
`not news_and_events_data` -> `news_and_events_data is not None`

rca/home/models.py
```diff
@@ -174,6 +177,32 @@ def clean(self):
...
+            try:
+                news_and_events_data = content.pull_news_and_events()
+            except content.CantPullFromRcaApi:
+                pass
```
tm-kn: Do we want to return `[]` here?

kevinhowbrook: Good point, thanks

rca/home/models.py
```diff
@@ -174,6 +177,32 @@ def clean(self):
...
+        cache_key = "latest_alumni_stories"
+        stories_data = cache.get(cache_key)
+        if stories_data is None:
+            try:
+                stories_data = content.pull_alumni_stories()
+            except content.CantPullFromRcaApi:
+                return []
+            else:
+                cache.set(cache_key, stories_data, settings.API_CONTENT_CACHE_TIMEOUT)
+            return stories_data
```
tm-kn: one indentation too much. You want to return this outside of the if-statement.

rca/api_content/content.py
```diff
@@ -0,0 +1,148 @@
...
+        if type == "Alumni stories":
+            detail = item["meta"]["detail_url"] + "?fields=_,feed_image,intro"
+
+        try:
+            response = requests.get(url=detail)
+            response.raise_for_status()
+        except requests.exceptions.HTTPError:
+            logger.exception("Error occured when fetching further detail data")
+
+        data = response.json()
```
tm-kn: You can't do this. What if exception was raised? Do you still care about the JSON content? You should raise the `CantPullFromRcaApiError` in the `except` block probably.

tm-kn: If, for whatever reason, you want the loop to continue regardless, you can also just `continue` within the except block, right?

kevinhowbrook:
(approved over slack)

PR #185, opened 2019-09-09
[119] Override basic auth get_client_ip function

kaedroho:
So we always check against the real IP rather than X-Forwarded-For. This is so we can verify that requests went via the old site rather than direct.


kevinhowbrook:
Thanks Karl!

PR #212, opened 2019-09-13
Course meta title - add degree level

nicklee:


nicklee:
No longer needed - being handled by wagtail promote tab

PR #253, opened 2019-10-04
178 gallery arrows

kevinhowbrook:
[Ticket 178](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/178)


kevinhowbrook:
> Should the conditional be around the whole component instead of just the arrows? seems like we shouldn't be outputting anything if there are no slides?

Good point, I've changed that

PR #258, opened 2019-11-13
Navigation updates: allow links to be relative

kevinhowbrook:


rca/navigation/models.py
```diff
@@ -7,8 +9,50 @@
...
+
+
+class URLOrRelativeURLFormField(forms.URLField):
+    default_validators = [url_or_relative_url_validator]
+
+    def to_python(self, value):
+        value = super().to_python(value)
+        if value:
+            if value.startswith("http:///"):
+                value = value[7:]
```
tm-kn: What is this? I have no idea what you are doing here.

kevinhowbrook: Without this, the validation won't pass, the return value for `/schools/` would be `http://schools/` so we need to strip the http from it to return `/schools/`

rca/navigation/models.py
```diff
@@ -7,8 +9,50 @@
 from wagtail.core.fields import StreamField
 
 
+def url_or_relative_url_validator(value):
+    if value.startswith("/"):
+        value = "http://rca.ac.uk" + value
```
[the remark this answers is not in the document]
kevinhowbrook: Due to the proxying setup, the primary host will remain rca-production.herokuapp.com so I think I need to leave this, but yea it's just to please the validator so it's a bit moot

rca/navigation/models.py
```diff
@@ -7,8 +9,50 @@
 from wagtail.core.fields import StreamField
 
 
+def url_or_relative_url_validator(value):
+    if value.startswith("/"):
```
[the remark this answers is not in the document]
kevinhowbrook: Thanks, I've implemented your suggestion

rca/navigation/models.py
tm-kn: IMO you could rely more on the URL manipulation capabilities of the standard library.

If it's `http:///schools/`, then this should work. Won't work for `http://schools/`.

```python
from urllib.parse import urlsplit, urlunsplit
    def to_python(self, value):
        value = super().to_python(value)
        if value:
            url_fields = list(urlsplit(value))
            # If netloc (domain) is empty, delete the scheme
            if not url_fields[1]:
                url_fields[0] = ''
             value = urlunsplit(url_fields)
```

Reference:
https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlsplit
https://github.com/django/django/blob/ca0d50f34ad14be2eca3067011c659ef0e66a0bc/django/forms/fields.py#L665-L694

rca/navigation/models.py
```diff
@@ -7,8 +11,50 @@
 from wagtail.core.fields import StreamField
 
 
+def url_or_relative_url_validator(value):
+    if not urlparse(value).netloc:
+        value = urljoin("https://rca.ac.uk/", value)
```
tm-kn: I'd add the comment that this domain is just for validation-sake and it won't actually be used anywhere, I hope I am right?

rca/navigation/models.py
kevinhowbrook: :+1: thanks Tomasz

PR #261, opened 2019-12-19
#254 Implement field spec for programmes index page

thibaudcolas:
See https://projects.torchbox.com/projects/rca-website-rebuild/tickets/254.

The `ProgrammeIndexPage` was already there as a placeholder so I only had to update its fields, and did the corresponding changes to the templates (still lots more to do there, I only did enough for the fields to be used).


rca/project_styleguide/templates/patterns/pages/programmes/programme_index.html
```diff
@@ -2,42 +2,59 @@
...
+        <input type="search" placeholder="{{ page.search_placeholder_text }}"/>
+
+        <div class="page__content">
+
+            {% if subpages %}
+                <div class="grid">
+                    {% if subpages.object_list.exists %}
+                        {% for subpage in subpages.object_list.specific %}
+                            <a class="card-with-image" href="{% pageurl subpage %}">
+                                {% if subpage.hero_image %}
```
kevinhowbrook: Do we need to work in the listing_image here? Usually, I use the listing image if specified, but if not use the hero

rca/project_styleguide/templates/patterns/pages/programmes/programme_index.yaml
```diff
@@ -1,26 +1,27 @@
...
+    title: Our Programmes
+    introduction: <p>Experience the world&apos;s <a href="https://www.example.com/">number one</a> university for art and design.</p>
+    search_placeholder_text: What are you interested in?
   subpages:
     object_list:
       exists: true
       specific:
         - title: Test sub page
-          listing_image: fake
+          hero_image: fake
```
kevinhowbrook: If we _do_ work in the listing_image it's worth changing one of these to use it so we can see it working in the pattern guide.

rca/project_styleguide/templates/patterns/pages/programmes/programme_index.html
thibaudcolas: According to https://projects.torchbox.com/projects/rca-website-rebuild/tickets/249 we’re meant to use the hero image for each programme, although I haven’t checked whether all programmes do have a hero image defined.

This will only be relevant when fetching this data from the API though (this listing in the template was just a placeholder).

PR #262, opened 2019-12-24
#255 Search results API

thibaudcolas:
See #264.


thibaudcolas:
Thanks @kevinhowbrook! I just realised I had made this PR against the wrong target branch, so will merge #264 with the same changes and the correct target now.

PR #268, opened 2020-01-02
Switch to eslint-config-torchbox

thibaudcolas:
I have a fair amount of JS code to write for the programmes listing so thought it would be a good idea to update the project’s ESlint config to https://github.com/torchbox/eslint-config-torchbox first.

I only did safe transforms and otherwise disabled rules that reported problems to keep this easy to review and merge – but there’s still lots of changes, so might be easier to review commit-by-commit.


.eslintrc.js
```diff
@@ -1,27 +1,12 @@
+const legacyCode = {
+    'class-methods-use-this': 0,
```
thibaudcolas: There were too many places using this code style for me to add individual "disable" comments. I don’t think having this rule disabled is an issue at all.

docs/front-end-tooling.md
```diff
@@ -7,18 +7,6 @@ This set of tooling should form the basis for any new wagtail projects. It can a
 You can run the tooling within the VM where Node.js is pre-installed, but if you are using Mac OS, you will likely have issues with performance of `npm install` and other `npm` commands. It is adviced to Mac OS users to have node on the host machine.
 To install node on the host machine we recommend using [`nvm`](https://github.com/creationix/nvm). Once you have `nvm` installed simply run `nvm install` to install and activate the version of node required for the project. Refer to the [`nvm` documentation](https://github.com/creationix/nvm#usage) for more details about available commands.
 
-## Setting up a new project from the wagtail-kit tooling
```
thibaudcolas: This and below – removing boilerplate from wagtail-kit.

PR #269, opened 2020-01-02
Feature/programmes listing fuzzy matching

kevinhowbrook:
Summary:

There is to be a search for page models using a text input which will work with a few fields (TBD for now, we know a few). We will need the ability for the user to enter a few words and have the search find suitable results. So searching for `Texti` will return `Textiles` results

At the moment, degree level and programme type ar supported so test with 

http://0.0.0.0:8000/api/v2/pages/?search=diplom
http://0.0.0.0:8000/api/v2/pages/?search=MRes

### DIFF
--- a/rca/programmes/models.py
+++ b/rca/programmes/models.py
@@ -538,6 +538,17 @@ class ProgrammePage(BasePage):
                 index.AutocompleteField("title", partial_match=True),
             ],
         ),
+        index.SearchField("programme_description_subtitle", partial_match=True),
+        index.AutocompleteField("programme_description_subtitle", partial_match=True),
+        index.SearchField("pathway_blocks", partial_match=True),
+        index.AutocompleteField("pathway_blocks", partial_match=True),
+        index.RelatedFields(
+            "degree_level",
+            [
+                index.SearchField("title", partial_match=True),
+                index.AutocompleteField("title", partial_match=True),
+            ],
+        ),
     ]
 
     api_fields = [
--- a/rca/programmes/models.py
+++ b/rca/programmes/models.py
@@ -531,13 +531,6 @@ class ProgrammePage(BasePage):
                 )
             ],
         ),
-        index.RelatedFields(
-            "degree_level",
-            [
-                index.SearchField("title", partial_match=True),
-                index.AutocompleteField("title", partial_match=True),
-            ],
-        ),
         index.SearchField("programme_description_subtitle", partial_match=True),
         index.AutocompleteField("programme_description_subtitle", partial_match=True),
         index.SearchField("pathway_blocks", partial_match=True),
--- a/rca/programmes/models.py
+++ b/rca/programmes/models.py
@@ -518,7 +518,27 @@ class ProgrammePage(BasePage):
         ]
     )
 
-    search_fields = BasePage.search_fields + []
+    search_fields = BasePage.search_fields + [
+        index.RelatedFields(
+            "programme_type",
+            [
+                index.RelatedFields(
+                    "programme_type",
+                    [
+                        index.SearchField("display_name", partial_match=True),
+                        index.AutocompleteField("display_name", partial_match=True),
+                    ],
+                )
+            ],
+        ),
+        index.RelatedFields(
+            "degree_level",
+            [
+                index.SearchField("title", partial_match=True),
+                index.AutocompleteField("title", partial_match=True),
+            ],
+        ),
+    ]
 
     api_fields = [
         APIField("name"),
--- a/rca/wagtailapi/endpoints.py
+++ b/rca/wagtailapi/endpoints.py
@@ -3,7 +3,6 @@ from wagtail.api.v2.filters import (
     FieldsFilter,
     RestrictedChildOfFilter,
     RestrictedDescendantOfFilter,
-    SearchFilter,
 )
 from wagtail.api.v2.serializers import PageSerializer
 
@@ -18,5 +17,5 @@ class PagesAPIEndpoint(endpoints.PagesAPIEndpoint):
         FieldsFilter,
         RestrictedChildOfFilter,
         RestrictedDescendantOfFilter,
-        SearchFilter,
+        filters.SearchFilter,
     ]
--- a/rca/wagtailapi/filters.py
+++ b/rca/wagtailapi/filters.py
@@ -1,6 +1,5 @@
 from django.conf import settings
 from rest_framework import filters
-from rest_framework.filters import BaseFilterBackend
 from wagtail.api.v2.utils import BadRequestError
 from wagtail.search.backends import get_search_backend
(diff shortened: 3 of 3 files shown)


rca/programmes/models.py
```diff
@@ -518,7 +518,27 @@ class ProgrammePage(BasePage):
...
+            ],
+        ),
+        index.RelatedFields(
+            "degree_level",
+            [
+                index.SearchField("title", partial_match=True),
+                index.AutocompleteField("title", partial_match=True),
+            ],
+        ),
+    ]
```
thibaudcolas: We probably also want the other fields displayed in the results to be added here, so the `programme_description_subtitle` and (if possible?) the pathway blocks’ label?.

rca/wagtailapi/filters.py
```diff
@@ -13,3 +17,50 @@ def filter_queryset(self, request, queryset, view):
             queryset = queryset.filter(degree_level__in=pks)
 
         return queryset
+
+
+class SearchFilter(filters.SearchFilter):
+    def filter_queryset(self, request, queryset, view):
+        """
+        This ovverides the wagtail core api filters.SearchFilter
```
thibaudcolas:
```suggestion
        This overrides the wagtail core api filters.SearchFilter
```

rca/programmes/models.py
kevinhowbrook: Which one is the pathways block label?

image

Assuming it's `Extend and develop...`?

kevinhowbrook: I've indexed the whole block as it's a streamfield, not an FK rel, ie all one blob in the db.

rca/wagtailapi/filters.py
```diff
@@ -13,3 +17,50 @@ def filter_queryset(self, request, queryset, view):
...
+                    "filtering by tag with a search query is not supported"
+                )
+
+            search_query = request.GET["search"]
+            search_operator = request.GET.get("search_operator", None)
+            order_by_relevance = "order" not in request.GET
+
+            sb = get_search_backend()
+            try:
+                queryset = sb.autocomplete(
```
tm-kn: This is not a concern, but I wonder if this will only search field indexed with `index.AutocompleteField` and `index.SearchField`.

rca/programmes/models.py
```diff
@@ -518,7 +518,38 @@ class ProgrammePage(BasePage):
...
+                index.SearchField("title", partial_match=True),
+                index.AutocompleteField("title", partial_match=True),
+            ],
+        ),
+        index.SearchField("programme_description_subtitle", partial_match=True),
+        index.AutocompleteField("programme_description_subtitle", partial_match=True),
+        index.SearchField("pathway_blocks", partial_match=True),
+        index.AutocompleteField("pathway_blocks", partial_match=True),
+        index.RelatedFields(
+            "degree_level",
```
thibaudcolas: Duplicate of the above?

rca/programmes/models.py
```diff
@@ -518,7 +518,38 @@ class ProgrammePage(BasePage):
...
+            [
+                index.SearchField("title", partial_match=True),
+                index.AutocompleteField("title", partial_match=True),
+            ],
+        ),
+        index.SearchField("programme_description_subtitle", partial_match=True),
+        index.AutocompleteField("programme_description_subtitle", partial_match=True),
+        index.SearchField("pathway_blocks", partial_match=True),
+        index.AutocompleteField("pathway_blocks", partial_match=True),
```
thibaudcolas: I’m not getting results with words from the pathways, I think this needs further work. Try for example "future", or "sound".

rca/programmes/models.py
```diff
@@ -518,7 +518,38 @@ class ProgrammePage(BasePage):
...
-    search_fields = BasePage.search_fields + []
+    search_fields = BasePage.search_fields + [
+        index.RelatedFields(
+            "programme_type",
+            [
+                index.RelatedFields(
+                    "programme_type",
+                    [
+                        index.SearchField("display_name", partial_match=True),
+                        index.AutocompleteField("display_name", partial_match=True),
```
thibaudcolas: Not getting results based on this - for example "Graduate".

rca/programmes/models.py
tm-kn: I wonder if adding this to the index will increase database size and whether it's a cost concern on Heroku.

rca/wagtailapi/filters.py
tm-kn: I am not 100% certain that you need to use `autocomplete` method.

Seems like the generic search method may be able to search through the "auto complete" index too - https://github.com/wagtail/wagtail/blob/28c104ff63e04f1ea4e440a789350f12e79ddc42/wagtail/contrib/postgres_search/backend.py#L302-L304 - but I may be interpreting this wrongly so worth giving it a test.

rca/wagtailapi/endpoints.py
```diff
@@ -18,5 +17,5 @@ class PagesAPIEndpoint(endpoints.PagesAPIEndpoint):
         FieldsFilter,
         RestrictedChildOfFilter,
         RestrictedDescendantOfFilter,
-        SearchFilter,
+        filters.SearchFilter,
```
tm-kn: Are there other pages that rely on API changes that we may be potentially breaking?

rca/wagtailapi/filters.py
kevinhowbrook: I'm pretty certain, testing a search for `diploma` using 3 letters `dip`:

without sb.autocomplete:

[!img](https://i.imgur.com/pKAkoNM.png)

with sb.autocomplete:

[!img](https://i.imgur.com/SQb53Xz.png)

rca/wagtailapi/endpoints.py
thibaudcolas: Nope

thibaudcolas:
My bad re the lack of results – I had just forgotten to update the index after pulling the branch. Thanks @kevinhowbrook for pointing it out.

rca/wagtailapi/endpoints.py
kevinhowbrook: At this moment no - we will be exposing pretty much all pages we create in the future though

rca/programmes/models.py
kevinhowbrook: resolved in https://github.com/torchbox/rca-wagtail-2019/pull/269/commits/c8fb3d00f4cd8dede916989a4b8587f3d683acee

rca/programmes/models.py
kevinhowbrook: Is the concern that we might be indexing twice?

thibaudcolas:
I need this in staging and will have a much easier time merging dependent work if merging this now – so merge merge merge. This won’t go live just yet so we can do more on this as part of the other tickets.

PR #273, opened 2020-01-03
#248 Programmes search form

thibaudcolas:
See https://projects.torchbox.com/projects/rca-website-rebuild/tickets/248. This implements the search form’s UI as well as logic needed to talk to the API and retrieve programme results, and a minimal UI for programme results.

Designs: https://www.figma.com/file/EWOmjORgGoHnfba7OSGmQ3ic/RCA---Phase-1-Shared?node-id=2099%3A130

Visually this is very similar to the site’s existing search form / field, so I chose to extend this code to hopefully make it easier to update the main site search to work similarly to this.

---

There is more work needed for the search to behave exactly as we want but that will be in a separate ticket.


rca/static_src/javascript/programmes/components/ProgrammeTeaser.js
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+import { programmePageShape } from '../programmes.types';
+
+/**
+ * A programme’s teaser info, to be displayed as part of search results.
+ */
+const ProgrammeTeaser = ({ programme }) => {
```
thibaudcolas: This is a placeholder just for the form to be tested meaningfully – not worth reviewing now. Same for the corresponding styles.

rca/static_src/javascript/programmes/components/ProgrammesResults.js
```diff
@@ -0,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+import { programmePageShape } from '../programmes.types';
+
+import ProgrammeTeaser from './ProgrammeTeaser';
+
+const ProgrammesResults = ({ programmes, hasActiveSearch }) => {
```
thibaudcolas: This is a placeholder just for the form to be tested meaningfully – not worth reviewing now.

rca/static_src/javascript/programmes/components/SearchForm.js
```diff
@@ -0,0 +1,128 @@
...
+                            }
+                        }}
+                    />
+                    {showClearButton ? (
+                        <button
+                            className="search__button button body body--two"
+                            type="button"
+                            onClick={(e) => {
+                                e.preventDefault();
+                                clearQuery();
```
thibaudcolas: I think the behaviour of what that "clearQuery" does might change, but not the form UI.

rca/static_src/javascript/programmes/programmes.api.js
```diff
@@ -0,0 +1,81 @@
...
+        limit: limit || null,
+        fields: fields.join(',') || null,
+        search: search || null,
+        ...filters,
+    };
+    Object.keys(parameters).forEach((param) => {
+        if (parameters[param] === null) {
+            delete parameters[param];
+        }
+    });
```
thibaudcolas: This is probably a bit too smart just for the one API client, but I’m hoping we can reuse this on other projects in the future.

rca/static_src/sass/abstracts/_mixins.scss
```diff
@@ -148,12 +148,20 @@
...
+}
+
+@mixin underline-hover-target-reset {
+    &::after {
+        width: 100%;
+        left: 0;
+    }
+}
+
+@mixin underline-hover() {
```
thibaudcolas: This might be a bit hard to review but I basically just split this existing mixin into two separate mixins, so I could reuse the same effect but in the case where we want the whole block to be hover-able, but only one of its components to show the effect.

rca/static_src/sass/components/_search.scss
```diff
@@ -23,20 +23,42 @@
...
         @include media-query(medium) {
-            top: 18px;
+            padding-top: 18px;
+            padding-bottom: 18px;
         }
     }
 
     &__icon {
-        fill: $color--white;
+        fill: currentColor;
```
thibaudcolas: This and the above changes should have no impact on the existing search form.

rca/static_src/sass/components/_search.scss
```diff
@@ -23,20 +23,42 @@
...
+        grid-column: 1 / span 2;
+        padding-bottom: $gutter * 2.5;
+
+        @include media-query(medium) {
+            padding-bottom: $gutter * 5.5;
+        }
+
+        @include media-query(large) {
+            grid-column: 2 / span 2;
+            max-width: 460px;
```
thibaudcolas: Couldn’t find a variable for this – elsewhere in the code max-width is also defined inline.

rca/static_src/sass/abstracts/_mixins.scss
chris-lawton: Yup, that makes sense

PR #274, opened 2020-01-03
Render out taxonomy items to json for react filters

kevinhowbrook:
Renders out taxonomy values to the template as json

### DIFF
--- a/rca/programmes/models.py
+++ b/rca/programmes/models.py
@@ -708,16 +708,20 @@ class ProgrammeIndexPage(BasePage):
             subpages = paginator.page(paginator.num_pages)
 
         # Listing filters
-        degree_levels = [i.title for i in DegreeLevel.objects.all()]
+        degree_levels = [
+            {"title": i.title, "id": i.id} for i in DegreeLevel.objects.all()
+        ]
         degree_levels_title = DegreeLevel._meta.verbose_name.capitalize()
-        programme_types = [i.display_name for i in ProgrammeType.objects.all()]
+        programme_types = [
+            {"title": i.display_name, "id": i.id} for i in ProgrammeType.objects.all()
+        ]
         programme_types_title = ProgrammeType._meta.verbose_name.capitalize()
         # TODO schools and subject
 
         filters = [
             {"title": degree_levels_title, "items": degree_levels},
             {"title": programme_types_title, "items": programme_types},
-            {"title": "Subject", "items": ["T", "B", "D"]},
+            {"title": "Subject", "items": [{"title": "TBD", "id": "100"}]},
         ]
 
         context.update(filters=filters)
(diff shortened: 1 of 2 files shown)


rca/project_styleguide/templates/patterns/pages/programmes/programme_index.html
```diff
@@ -20,6 +20,17 @@ <h1 class="title-area__header heading heading--display-two">{{ title_multiline|l
...
+            <div class="filters">
+                {% for f in filters %}
+                    <ul class="filters filters__list">
+                        {{ f.title }}
+                        {% for i in f.items %}
+                            <li class="filters filters__list item">{{ i }}</li>
+                        {% endfor %}
+                    </ul>
+                {% endfor %}
+            </div>
```
kevinhowbrook: @thibaudcolas feel free to remove this markup, just there for testing purposes really and as an example

rca/programmes/models.py
```diff
@@ -707,6 +707,20 @@ def get_context(self, request, *args, **kwargs):
         except EmptyPage:
             subpages = paginator.page(paginator.num_pages)
 
+        # Listing filters
+        degree_levels = [i.title for i in DegreeLevel.objects.all()]
+        degree_levels_title = DegreeLevel._meta.verbose_name.capitalize()
```
thibaudcolas: Oh that’s pretty cool, didn’t think we could do that!

rca/programmes/models.py
```diff
@@ -707,6 +707,20 @@ def get_context(self, request, *args, **kwargs):
         except EmptyPage:
             subpages = paginator.page(paginator.num_pages)
 
+        # Listing filters
+        degree_levels = [i.title for i in DegreeLevel.objects.all()]
```
thibaudcolas: I would also need the key/id for each item so they can be passed over to the API to apply the correct filter.

rca/programmes/models.py
```diff
@@ -707,6 +707,20 @@ def get_context(self, request, *args, **kwargs):
...
+        degree_levels_title = DegreeLevel._meta.verbose_name.capitalize()
+        programme_types = [i.display_name for i in ProgrammeType.objects.all()]
+        programme_types_title = ProgrammeType._meta.verbose_name.capitalize()
+        # TODO schools and subject
+
+        filters = [
+            {"title": degree_levels_title, "items": degree_levels},
+            {"title": programme_types_title, "items": programme_types},
+            {"title": "Subject", "items": ["T", "B", "D"]},
```
thibaudcolas: 😄

rca/project_styleguide/templates/patterns/pages/programmes/programme_index.html
thibaudcolas: 👍 good idea

rca/programmes/models.py
kevinhowbrook: Yea - it's unlikely to change, but if it does, at least we only change it in one place

rca/programmes/models.py
kevinhowbrook: added iID in https://github.com/torchbox/rca-wagtail-2019/pull/274/commits/477264c4c7826f47a1b393ef9f6577bedce46035

PR #277, opened 2020-01-07
Don't use a new database connection for every request

kevinhowbrook:


kevinhowbrook:
approved here https://github.com/torchbox/rca-wagtail-2019/pull/276

PR #278, opened 2020-01-08
#253 Programme category views

thibaudcolas:
See https://projects.torchbox.com/projects/rca-website-rebuild/tickets/253.

Design: https://www.figma.com/file/EWOmjORgGoHnfba7OSGmQ3ic/RCA---Phase-1-Shared?node-id=1810%3A2

---

The categories do not match what is described on the ticket and in the designs but otherwise this works the same. Ordering of the items is also to be done separately.

PR #279, opened 2020-01-08
#250 Programmes filter by category results

thibaudcolas:
Depends on #278. There isn’t many changes here so this will be much easier to review once #278 is merged. See https://projects.torchbox.com/projects/rca-website-rebuild/tickets/250.

Design: https://www.figma.com/file/EWOmjORgGoHnfba7OSGmQ3ic/RCA---Phase-1-Shared?node-id=1810%3A2

---

This reuses the programmes’ search existing results view, with a light theme, to display results from categories.

There is still more work to be done for the transitions between those different views in a separate ticket.


chris-lawton:
Looks good @thibaudcolas 👍

PR #289, opened 2020-01-13
Feature/programmes listing back end filters

kevinhowbrook:
## Summary

The new programme listing pages will feature listing programme pages with the filters, School, Subject and Type. The PR adds the needed taxonomy models and page models along  with logic to esnure the new models can be used to filter API requests. 

[Ticket](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/256)

- [x] Remove values, and legacy_slug field from programme type and add the new 'Type' values. 
- [x] Create new 'Subject' taxonomy with description and title fields. 
- [x] Add relationship for 'Subject' taxonomy > ProgrammePage
- [x] Create Page model for Schools and Research Centers (SchoolsAndResearchPage)
- [x] Add relationship for for ProgrammePage > SchoolsAndResearchPage
- [x] Expose 'Programme Type' as a filter for the listing page.
- [x] Expose 'Subject' Taxonomy as a filter for the listing page.
- [x] Expose 'SchoolsAndResearchPage' Page relationship as a filter for the listing page.
- [x] Ensure the schools and research page cannot be viewed on the front end > Managed with page permission visibility
- [x] Subjects taxonomy should support more than one value
- [x] Ensure the search filter is searching all the taxonomies (multiple subjects for example)


rca/wagtailapi/filters.py
```diff
@@ -19,6 +22,43 @@ def filter_queryset(self, request, queryset, view):
...
+                queryset = queryset.model.objects.filter(
+                    subjects__subject__in=subject_ids
+                )
+
+        return queryset
+
+
+class RelatedSchoolsFilter(filters.BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        if hasattr(queryset.model, "related_schools_and_research_pages"):
```
tm-kn: Again, you may want to check it's a Django field.

rca/wagtailapi/filters.py
```diff
@@ -19,6 +22,43 @@ def filter_queryset(self, request, queryset, view):
...
+            school_ids = request.GET.getlist("related_schools_and_research_pages", [])
```
tm-kn: Sanitise the input, e.g. `[int(id) for id in request.GET.getlist("related_schools_and_research_pages", [])]`

rca/wagtailapi/filters.py
```diff
@@ -19,6 +22,43 @@ def filter_queryset(self, request, queryset, view):
...
+            if school_ids:
+                school_pages = (
+                    SchoolsAndResearchPage.objects.all()
```
tm-kn: `all` is redundant here. You can delete it. Do you not want to use `public()` too or check the permissions of the user?

rca/wagtailapi/filters.py
```diff
@@ -19,6 +22,43 @@ def filter_queryset(self, request, queryset, view):
...
+                    .live()
+                    .filter(id__in=school_ids)
+                )
+                programme_pages_with_school_pages_relationship = (
```
tm-kn: Maybe try to shorten this variable name, e.g. `programmes_to_schools_qs`?

rca/wagtailapi/filters.py
```diff
@@ -19,6 +22,43 @@ def filter_queryset(self, request, queryset, view):
...
+                    ProgrammePageRelatedSchoolsAndResearchPage.objects.all()
+                )
+                relationship_objects = programme_pages_with_school_pages_relationship.filter(
+                    page__in=school_pages
```
tm-kn: Explicit is better - `page_id__in`

rca/programmes/serializers.py
```diff
@@ -7,3 +8,28 @@ class DegreeLevelSerializer(serializers.ModelSerializer):
...
+            for v in value.all():
+                subject = {}
+                subject["id"] = v.subject.id
+                subject["title"] = v.subject.title
+                subject["description"] = v.subject.description
+                subjects.append(subject)
+        return subjects
+
+
+class RelatedSchoolSerializer(Field):
```
tm-kn: This could be a model serializer.

rca/wagtailapi/filters.py
```diff
@@ -19,6 +22,43 @@ def filter_queryset(self, request, queryset, view):
...
+                ).values_list(
+                    "id", flat=True
+                )
+
+                queryset = queryset.model.objects.filter(
+                    related_schools_and_research_pages__id__in=relationship_objects
```
tm-kn: `related_schools_and_research_pages__page_id__in=school_pages.values_list('pk', flat=True)`

Wouldn't this make the `relationship_objects` queryset redundant?

rca/programmes/serializers.py
```diff
@@ -7,3 +8,28 @@ class DegreeLevelSerializer(serializers.ModelSerializer):
     class Meta:
         model = DegreeLevel
         fields = ("id", "title")
+
+
+class RelatedSubjectSerializer(Field):
```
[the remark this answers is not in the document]
kevinhowbrook: I'd like to use ModelSerializer, but the relationship is proving to be hard figure out. The relationship is set so there can be more than one subject. EG Subject --> Placement Model --> ProgrammePage model.
I've tried configuring a nested serializer with no luck though

```

class SubjectSerializer(serializers.ModelSerializer):
    class Meta:
        model = Subject
        fields = ('id', 'title')

class RelatedSubjectSerializer(serializers.ModelSerializer):
    subject = SubjectSerializer(many=True)
    class Meta:
        model = ProgrammePageSubjectPlacement
        fields = ("subject",)
```

rca/wagtailapi/filters.py
kevinhowbrook: Yep - I'll remove `all()` I need to avoid using public though as the one schools and research page that will exist will not be available as public, but _will_ need to be used as a filter

rca/wagtailapi/filters.py
kevinhowbrook: yep you are right... I'll amend this

rca/programmes/serializers.py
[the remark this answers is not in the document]
tm-kn: >     subject = SubjectSerializer(many=True)

Why `many`? It's only a ForeignKey, so it will return one instance, right?

rca/programmes/serializers.py
tm-kn: `APIField("subjects", serializer=subject_serializer(many=True)),` This is a list, right?

rca/programmes/serializers.py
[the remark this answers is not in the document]
kevinhowbrook: I've removed these serializers for now. They aren't necessarily needed for this feature

rca/programmes/serializers.py
kevinhowbrook: removed

PR #294, opened 2020-01-20
Adds a manual cache clearing mechanism for the api content feeds

kevinhowbrook:
### Context
The homepage will feature content pulled in from the legacy API. The pulled content is [stored in the cache for 24h](https://github.com/torchbox/rca-wagtail-2019/blob/staging/rca/settings/base.py#L686)

### Changes
There is a requirement to force clear this cache so this PR adds a new section in the admin UI where this can be done. If this is implemented I will recommend we set the cache on production for the pulled content to be larger than 24h.

### DIFF
--- a/rca/api_content/templates/api_content/api_content_admin.html
+++ b/rca/api_content/templates/api_content/api_content_admin.html
@@ -1,0 +1,32 @@
+{% extends "wagtailadmin/base.html" %}
+{% load wagtailadmin_tags static i18n %}
+{% block titletag %}{% trans "Dashboard" %}{% endblock %}
+{% block bodyclass %}homepage{% endblock %}
+
+
+{% block content %}
+    <script>
+    function clear_cache() {
+        $.get('/admin/api-cache/clear/', function (data) {
+            $("#cache_clear_button").removeClass("button-longrunning-active")
+            $("#cache_clear_button").prop("disabled", false);
+            $('.content').prepend('<div class="messages"><ul><li class="success">Success! API feed cache has been cleared</li></ul></div>')
+        });
+    }
+    </script>
+    <header class="merged nice-padding">
+        <div class="row row-flush">
+            <div class="col12">
+                <h1>Content feed settings</h1>
+            </div>
+        </div>
+    </header>
+    <div class="nice-padding">
+        <div class="help-block help-info">
+        <p>Some pages on the webstie use a content feed from the legacy API, for example, the News and Events section on the bottom of the homepage . This data is stored in the cache and automatically expires after one day.</p>
+        <p>You can manually clear the cache of the content feed here, this will cause new content to be pulled from the legacy API</p>
+        </div>
+        <button id="cache_clear_button" class="button button-longrunning" onclick="clear_cache()" id="button-arbitrarily-bigger"><span class="icon icon-spinner"></span>Clear API content feed cache</button>
+    </div>
+
+{% endblock %}
--- a/rca/api_content/views.py
+++ b/rca/api_content/views.py
@@ -1,0 +1,13 @@
+from django.core.cache import cache
+from django.http import HttpResponseRedirect
+from django.views.generic import TemplateView
+
+
+class ApiContentCacheSettings(TemplateView):
+    template_name = "api_content/api_content_admin.html"
+
+
+def ApiContentCacheClear(request):
+    cache_keys = ["latest_news_and_events", "latest_alumni_stories"]
+    cache.delete_many(cache_keys)
+    return HttpResponseRedirect("/admin/api-cache/")
--- a/rca/api_content/wagtail_hooks.py
+++ b/rca/api_content/wagtail_hooks.py
@@ -1,4 +1,7 @@
 from django.conf.urls import url
+from django.urls import reverse
+from django.utils.translation import ugettext_lazy as _
+from wagtail.admin.menu import MenuItem
 from wagtail.core import hooks
 
 from .views import ApiContentCacheClear, ApiContentCacheSettings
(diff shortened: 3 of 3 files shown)


rca/api_content/templates/api_content/api_content_admin.html
```diff
@@ -0,0 +1,32 @@
+{% extends "wagtailadmin/base.html" %}
+{% load wagtailadmin_tags static i18n %}
+{% block titletag %}{% trans "Dashboard" %}{% endblock %}
+{% block bodyclass %}homepage{% endblock %}
+
+
+{% block content %}
+    <script>
```
tm-kn: I do not quite understand why this requires JavaScript rather than just having a normal HTML form. Also, no CSRF check (not that this is crucial in this situation).

rca/api_content/templates/api_content/api_content_admin.html
```diff
@@ -0,0 +1,32 @@
...
+    function clear_cache() {
+        $.get('/admin/api-cache/clear/', function (data) {
```
tm-kn: Why not use `fetch`? https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

rca/api_content/templates/api_content/api_content_admin.html
```diff
@@ -0,0 +1,32 @@
...
+            $("#cache_clear_button").removeClass("button-longrunning-active")
```
tm-kn: No need to use jQuery IMO.

```js
async function clearCache() {
  try {
    const response = await fetch("/admin/api-cache/clear/");
    if (!response.ok) {
      throw Error(`Unexpected HTTP code ${response.status}`);
    }
    document.getElementById('cache_clear_button').classList.remove('button-longrunning-active');
    document.getElementById('cache_clear_button').disabled = true;
  } catch(e) {
    // error handling.
  }
}
```
https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

rca/api_content/templates/api_content/api_content_admin.html
```diff
@@ -0,0 +1,32 @@
...
+            $("#cache_clear_button").prop("disabled", false);
+            $('.content').prepend('<div class="messages"><ul><li class="success">Success! API feed cache has been cleared</li></ul></div>')
```
tm-kn: I think having a hidden div is better than dynamically generating HTML code in JS.

```js
document.getElementById('success-message').style.display = '';
```

rca/api_content/templates/api_content/api_content_admin.html
tm-kn: Also, no error handling here. That's why I think we should not bother with JS because Django provides everything out of the box.

rca/api_content/views.py
```diff
@@ -0,0 +1,13 @@
+from django.core.cache import cache
+from django.http import HttpResponseRedirect
+from django.views.generic import TemplateView
+
+
+class ApiContentCacheSettings(TemplateView):
+    template_name = "api_content/api_content_admin.html"
+
+
+def ApiContentCacheClear(request):
```
tm-kn: Why use camel case for function-based views? https://docs.djangoproject.com/en/3.0/topics/http/views/ Django docs do not seem to use this. Is that what Wagtail does?

rca/api_content/views.py
```diff
@@ -0,0 +1,13 @@
+from django.core.cache import cache
+from django.http import HttpResponseRedirect
+from django.views.generic import TemplateView
+
+
+class ApiContentCacheSettings(TemplateView):
+    template_name = "api_content/api_content_admin.html"
```
tm-kn: Out of curiosity, does this require user to actually be logged in to the admin?

rca/api_content/wagtail_hooks.py
```diff
@@ -0,0 +1,27 @@
...
+from .views import ApiContentCacheClear, ApiContentCacheSettings
+
+
+@hooks.register("register_admin_urls")
+def urlconf_time():
+    return [
+        url(
+            r"^api-cache/", ApiContentCacheSettings.as_view(), name="api-cache-settings"
+        ),
+        url(r"^api-cache/clear/", ApiContentCacheClear),
```
tm-kn: Why two URLs? Shouldn't it just be `GET` and `POST` on one URL?

rca/api_content/templates/api_content/api_content_admin.html
```diff
@@ -0,0 +1,32 @@
...
+                <h1>Content feed settings</h1>
+            </div>
+        </div>
+    </header>
+    <div class="nice-padding">
+        <div class="help-block help-info">
+        <p>Some pages on the webstie use a content feed from the legacy API, for example, the News and Events section on the bottom of the homepage . This data is stored in the cache and automatically expires after one day.</p>
+        <p>You can manually clear the cache of the content feed here, this will cause new content to be pulled from the legacy API</p>
+        </div>
+        <button id="cache_clear_button" class="button button-longrunning" onclick="clear_cache()" id="button-arbitrarily-bigger"><span class="icon icon-spinner"></span>Clear API content feed cache</button>
```
tm-kn: I'd go against using HTML to register JavaScript event handlers.

```js
document.getElementById('cache_clear_button').addEventListener('click', () => clear_cache());
```
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

tm-kn:
Also, couldn't we purge remotely from the verdant site with signal handlers? We could just establish a pre-shared token that has to match as a security measure.

rca/api_content/views.py
kevinhowbrook: no that's my error.... thanks

PR #311, opened 2020-02-07
Fix/listing pages

kevinhowbrook:
Adds arrow to contact URL link.
Removes back button then the search is used and no filter is aplied


rca/project_styleguide/templates/patterns/organisms/contact/contact.html
```diff
@@ -17,6 +17,7 @@ <h2 class="contact__heading heading heading--four">{{ heading }}</h2>
         {% else %}
             <a href="{{ page.contact_url }}" class="contact__link link link--secondary link--link body body--two" rel="noopener">
             <span class="link__label">Register your interest with us here</span>
+            <svg width="12" height="8" class="link__icon"><use xlink:href="#arrow"></use></svg>
```
thibaudcolas:
```suggestion
            <svg width="12" height="8" class="link__icon" aria-hidden="true"><use xlink:href="#arrow"></use></svg>
```

PR #313, opened 2020-03-11
Limit requests to the legacy site to stop the app crashing

kevinhowbrook:
If the verdant site we pull content from over the api is down, we should handle the timeout because heroku will crash at 30 seconds

### DIFF
--- a/rca/api_content/content.py
+++ b/rca/api_content/content.py
@@ -5,6 +5,7 @@ import requests
 from bs4 import BeautifulSoup
 from django.conf import settings
 from django.http.request import QueryDict
+from requests.exceptions import ConnectionError, HTTPError, Timeout
 
 """
 Static methods for adding content from the live RCA api
@@ -39,9 +40,37 @@ class CantPullFromRcaApi(Exception):
     pass
 
 
+def fetch_data(url):
+    try:
+        response = requests.get(url, timeout=10)
+        response.raise_for_status()
+    except Timeout:
+        logger.exception(f"Timeout error occurred when fetching data from {url}")
+        raise CantPullFromRcaApi(
+            "Error occured when fetching further detail data from {url}"
+        )
+    except (HTTPError, ConnectionError):
+        logger.exception(
+            "HTTP/ConnectionError occured when fetching further detail data from {url}"
+        )
+        raise CantPullFromRcaApi(
+            "Error occured when fetching further detail data from {url}"
+        )
+    except Exception:
+        logger.exception(
+            f"Exception occured when fetching further detail data from {url}"
+        )
+        raise CantPullFromRcaApi(
+            "Error occured when fetching further detail data from {url}"
+        )
+    else:
+        return response.json()
+
+
 def parse_items_to_list(data, type):
     items = []
-
+    if not data:
+        return []
     for item in data["items"]:
         _item = {}
         if type == "News":
@@ -58,18 +87,14 @@ def parse_items_to_list(data, type):
                 item["meta"]["detail_url"]
                 + "?fields=_,first_published_at,social_image,body"
             )
-        try:
-            response = requests.get(url=detail, timeout=5)
-            response.raise_for_status()
-        except requests.exceptions.HTTPError:
-            error_text = "Error occured when fetching further detail data"
-            logger.exception(error_text)
-            raise CantPullFromRcaApi(error_text)
 
-        data = response.json()
+        data = fetch_data(detail)
+        if not data:
+            return []
+
         if "social_image" in data and data["social_image"]:
             social_image = data["social_image"]["meta"]["detail_url"]
-            social_image = requests.get(url=social_image, timeout=5)
+            social_image = requests.get(url=social_image, timeout=10)
             social_image = social_image.json()
             if "url" in social_image["rca2019_feed_image"]:
                 social_image_url = social_image["rca2019_feed_image"]["url"]
@@ -77,7 +102,7 @@ def parse_items_to_list(data, type):
                 _item["image"] = social_image_url
                 _item["image_small"] = social_image_small_url
                 _item["image_alt"] = social_image["alt"]
-        date = False
+        date = None
         if type == "News":
             date = data["date"]
             _item["type"] = type
@@ -113,9 +138,12 @@ def parse_items_to_list(data, type):
 
 
 def pull_news_and_events(programme_type_slug=None):
-    # 'News' is actually a mixture of NewItem and RcaBlogPage models.
+    # 'News' is actually a mixture of NewsItem and RcaBlogPage models.
     # So pull 3 of both from the API and order them by the date field.
     # Then select how many we need depending on the events content.
+    NEWS_ITEMS = 3
+    # By default, work with 3 news items but if we pull events, get 2
+    # so we end up with 2 x News/Blog and 1 x Event.
 
     # First get the Events
     query = QueryDict(mutable=True)
@@ -134,25 +162,15 @@ def pull_news_and_events(programme_type_slug=None):
     query = query.urlencode()
     events_url = f"{settings.API_CONTENT_BASE_URL}/api/v2/pages/?{query}"
 
-    # By default, work with 3 news items but if we pull events, get 2
-    # so we end up with 2 x News/Blog and 1 x Event.
-    news_items_to_get = 3
-
     events_data = []
-    try:
-        response = requests.get(url=events_url, timeout=5)
-        response.raise_for_status()
-        logger.info("pulling Events from API")
-    except requests.exceptions.HTTPError:
-        error_text = "Error occured when fetching event data"
-        logger.exception(error_text)
-        raise CantPullFromRcaApi(error_text)
-    else:
-        data = response.json()
-
-        if data["meta"]["total_count"] > 0:
-            news_items_to_get = 2
-            events_data = parse_items_to_list(data, "Event")
+    fetched_event_data = fetch_data(events_url)
+    if (
+        fetched_event_data
+        and "total_count" in fetched_event_data["meta"]
+        and fetched_event_data["meta"]["total_count"] > 0
+    ):
+        NEWS_ITEMS = 2
+        events_data = parse_items_to_list(fetched_event_data, "Event")
 
     # News and Blogs
     # Pull 3 Blog items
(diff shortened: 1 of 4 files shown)


rca/api_content/content.py
```diff
@@ -59,8 +60,12 @@ def parse_items_to_list(data, type):
                 + "?fields=_,first_published_at,social_image,body"
             )
         try:
```
richbrennan: This `try/except` pattern is repeated a few times, could it be wrapped in a function?

kevinhowbrook: Thanks Rich, good idea. I've attempted this here https://github.com/torchbox/rca-wagtail-2019/pull/313/commits/02e690aa81a9c9317b9b473a0339a359cf03d57a.

I'm not 100% on this, mainly because if the execution makes it to here https://github.com/torchbox/rca-wagtail-2019/pull/313/commits/02e690aa81a9c9317b9b473a0339a359cf03d57a#diff-3a8951d04a021439a3fbc9866c9a2d16R155 then there will be no response defined; but I guess before that the exception would be raised? Should I be checking for the response at this point?

kevinhowbrook: Also, I've changed the timeout to be 5 seconds which is a bit more reasonable

richbrennan: RE the `try_and_fetch()` function, another repeated line is `data = response.json()`, which suggests that what your code really wants from the function is appropriate data, rather than a response that it then needs to deal with (and which it doesn't need/want to know about).

You could rename the function to be something like `fetch_data(url)` to encapsulate all the `requests` boilerplate, and maybe return `None` if it can't fetch it so your code can check.

kevinhowbrook: I've updated the code following your suggestions.
I've also tested this by replacing https://github.com/torchbox/rca-wagtail-2019/pull/313/files#diff-3a8951d04a021439a3fbc9866c9a2d16R48 with `data = []` to test the page still loads with no data.

Ideally, I need some tests for this, which I will add as a backlog ticket.

rca/api_content/content.py
```diff
@@ -53,11 +55,14 @@ def try_and_fetch(url):
         error_text = "Error occured when fetching further detail data"
         logger.exception(error_text)
         raise CantPullFromRcaApi(error_text)
+    else:
+        return None
```
richbrennan: When used in a `try` statement, `else` runs if an exception _isn't_ raised, so this line won't run. See https://geek-university.com/python/the-try-except-else-statements/

Its best to have as few statements in a `try` as possible, just the ones that might raise, then have the safer ones in the `else`.

So (assuming the function doesn't re-raise):

```
def fetch_data(url):
    logger.info(f"pulling data from API {url}")
    try:
        response = requests.get(url, timeout=5)
        response.raise_for_status()
    except Timeout:
        logger.exception(f"Timeout error occurred when fetching data from {url}")
    except (HTTPError, ConnectionError):
        logger.exception(f"Connection error occurred when fetching further detail data from {url}")
    else:
        return response.json()
```

rca/api_content/content.py
```diff
@@ -53,11 +55,14 @@ def try_and_fetch(url):
         error_text = "Error occured when fetching further detail data"
         logger.exception(error_text)
         raise CantPullFromRcaApi(error_text)
```
richbrennan: Rather than halting the program, would it be worth returning `None` here?

richbrennan: Same for above

rca/api_content/content.py
```diff
@@ -150,9 +156,8 @@ def pull_news_and_events(programme_type_slug=None):
     news_items_to_get = 3
 
     events_data = []
-    response = try_and_fetch(events_url)
-    data = response.json()
-    if data["meta"]["total_count"] > 0:
+    data = fetch_data(events_url)
+    if data and data["meta"]["total_count"] > 0:
```
richbrennan: Potential for index errors here?

rca/api_content/content.py
```diff
@@ -228,8 +231,7 @@ def pull_alumni_stories(programme_type_slug=None):
     query = query.urlencode()
     url = f"{settings.API_CONTENT_BASE_URL}/api/v2/pages/?{query}"
 
-    response = try_and_fetch(url)
-    data = response.json()
+    data = fetch_data(url)
     alumni_stories_standrad_page_data = parse_items_to_list(
```
richbrennan: Typo `alumni_stories_standrad_page_data`

rca/api_content/content.py
```diff
@@ -75,8 +80,9 @@ def parse_items_to_list(data, type):
                 + "?fields=_,first_published_at,social_image,body"
             )
 
-        response = try_and_fetch(detail)
-        data = response.json()
+        data = fetch_data(detail)
+        if not data:
+            return []
```
richbrennan: If there isn't data (maybe because the fetch raised an exception), will anything be shown on the FE?

richbrennan:
Does this code run on page load or as part of a management command?

rca/api_content/content.py
kevinhowbrook: > Does this code run on page load or as part of a management command?

Runs on page load, cache is regenerated after 15 mins though

kevinhowbrook:
> Does this code run on page load or as part of a management command?

Runs on page load, cache is regenerated after 15 mins though

rca/api_content/content.py
kevinhowbrook: It's all hidden if there is no data. I think there was a title left in, I'll ensure that's also hidden

rca/api_content/content.py
kevinhowbrook: I'm not sure I know the answer to this - what I can say though is that the `CantPullFromRcaApi` exception is set to pass so it shouldn't halt it?

rca/api_content/content.py
kevinhowbrook: Thanks, I've neatened this up a bit in https://github.com/torchbox/rca-wagtail-2019/pull/313/commits/bb89344e2606ee15594bdbad3ac506c5f7b4aae4

kevinhowbrook: I've added an exception for anything that is not specified in https://github.com/torchbox/rca-wagtail-2019/pull/313/commits/7b7f88507f886ead17917bb1faa5626ef1b8a16f

I've tested raising theses locally and making sure the page loads with no data should any of the exceptions get caught by raising them in the code at the `try` point. All loads ok

rca/api_content/content.py
kevinhowbrook: We covered this on a call, the general recommendation is that we refactor the model code so it's not try to catch this exception [here](https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/home/models.py#L186) - I'll roll this into another ticket due to time restraints

kevinhowbrook: I'll add it to https://projects.torchbox.com/projects/rca-website-rebuild/tickets/308

rca/api_content/content.py
kevinhowbrook: The answer to this was that the custom exception `CantPullFromRca` is caught here https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/home/models.py#L186

rca/api_content/content.py
```diff
@@ -171,17 +183,8 @@ def pull_news_and_events(programme_type_slug=None):
...
-        logger.info("Pulling Blogs from API")
-    except requests.exceptions.HTTPError:
-        error_text = "Error occured when fetching Blog data"
-        logger.exception(error_text)
-        raise CantPullFromRcaApi(error_text)
-    else:
-        data = response.json()
-        blog_data = parse_items_to_list(data, "News")
+    data = fetch_data(blog_url)
+    blog_data = parse_items_to_list(data, "News")
```
richbrennan, line 187: Are "news" and "blog" being used interchangeably? If so it'll help readability if just one term is used, otherwise a comment to explain might help

rca/api_content/content.py
```diff
@@ -39,9 +40,34 @@ class CantPullFromRcaApi(Exception):
...
+    except Exception:
+        logger.exception(
+            f"Exception occured when fetching further detail data from {url}"
+        )
+    else:
+        return response.json()
+
+
 def parse_items_to_list(data, type):
     items = []
```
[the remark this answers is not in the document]
kevinhowbrook: It's appended to here https://github.com/torchbox/rca-wagtail-2019/blob/7b7f88507f886ead17917bb1faa5626ef1b8a16f/rca/api_content/content.py#L132

rca/api_content/content.py
```diff
@@ -58,18 +84,14 @@ def parse_items_to_list(data, type):
...
-            response = requests.get(url=detail, timeout=5)
-            response.raise_for_status()
-        except requests.exceptions.HTTPError:
-            error_text = "Error occured when fetching further detail data"
-            logger.exception(error_text)
-            raise CantPullFromRcaApi(error_text)
-
-        data = response.json()
+
+        data = fetch_data(detail)
```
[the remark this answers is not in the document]
kevinhowbrook: Yea I need to name these better, will rename them. Thanks

rca/api_content/content.py
```diff
@@ -139,20 +161,10 @@ def pull_news_and_events(programme_type_slug=None):
     news_items_to_get = 3
```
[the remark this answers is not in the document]
kevinhowbrook: changed in https://github.com/torchbox/rca-wagtail-2019/pull/313/commits/8ae4a81e2b7ea045facc0fd0f60876928e17f9bc

rca/api_content/content.py
```diff
@@ -139,20 +161,10 @@ def pull_news_and_events(programme_type_slug=None):
...
-        raise CantPullFromRcaApi(error_text)
-    else:
-        data = response.json()
-
-        if data["meta"]["total_count"] > 0:
-            news_items_to_get = 2
-            events_data = parse_items_to_list(data, "Event")
+    data = fetch_data(events_url)
+    if data and "total_count" in data["meta"] and data["meta"]["total_count"] > 0:
+        news_items_to_get = 2
```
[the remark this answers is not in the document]
kevinhowbrook: changed in https://github.com/torchbox/rca-wagtail-2019/pull/313/commits/8ae4a81e2b7ea045facc0fd0f60876928e17f9bc

rca/api_content/content.py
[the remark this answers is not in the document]
richbrennan: Good spot! :-D

rca/api_content/content.py
```diff
@@ -134,25 +164,15 @@ def pull_news_and_events(programme_type_slug=None):
...
-        if data["meta"]["total_count"] > 0:
-            news_items_to_get = 2
-            events_data = parse_items_to_list(data, "Event")
+    fetched_event_data = fetch_data(events_url)
+    if (
+        fetched_event_data
+        and "total_count" in fetched_event_data["meta"]
+        and fetched_event_data["meta"]["total_count"] > 0
+    ):
+        news_items_to_get = NEWS_ITEMS_WITH_EVENTS
```
richbrennan, line 174: You shouldn't need to assign the constant to a variable (its just a normal variable anyway)

PR #315, opened 2020-03-11
Update revisions missing programme_type as it's causing viewing errors

kevinhowbrook:
We made some changes to programme types a while back which saw us delete all the existing taxonomy items, this has caused errors when viewing page revisions with dangling taxonomy links. 
This PR has a migration to update all the page revisions to have programme_type values

### DIFF
--- a/rca/programmes/migrations/0051_update_revisions.py
+++ b/rca/programmes/migrations/0051_update_revisions.py
@@ -1,0 +1,25 @@
+import json
+
+from django.core.serializers.json import DjangoJSONEncoder
+from django.db import migrations
+
+from rca.programmes.models import ProgrammePage
+
+
+def update_revisions_with_current_programme_type(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


rca/programmes/migrations/0051_update_revisions.py
```diff
@@ -0,0 +1,21 @@
+import json
+
+from django.core.serializers.json import DjangoJSONEncoder
+from django.db import migrations
+
+from rca.programmes.models import ProgrammePage
+
+
+def update_revisions_with_current_programme_type(apps, schema_editor):
+    for page in ProgrammePage.objects.all():
```
richbrennan: Models in migrations need to be imported from the app registry, like:
`ProgrammePage = apps.get_model('programmes', 'ProgrammePage')`.

See https://docs.djangoproject.com/en/3.0/topics/migrations/#data-migrations

PR #325, opened 2020-03-16
Feature/278 external link fe

nicklee:
External link styling using just css


rca/project_styleguide/templates/patterns/atoms/link-tertiary/link-tertiary--external.html
```diff
@@ -0,0 +1,4 @@
+<a href="{{ value.href }}" class="link link--tertiary link--link">
+    <span class="link__label">{{ value.text }}</span>
+    <svg width="12" height="8" class="link__icon"><use xlink:href="#arrow"></use></svg>
```
chris-lawton: should we follow our pattern of aria hiding this?

nicklee: merci beaucoup

PR #328, opened 2020-03-17
[WIP] Feature/staff detail FE

nicklee:
Based on `feature/short-course-ALL-FE` - with additional of staff detail template and components

### DIFF
--- a/rca/project_styleguide/templates/patterns/atoms/profile-image/profile-image.html
+++ b/rca/project_styleguide/templates/patterns/atoms/profile-image/profile-image.html
@@ -1,6 +1,6 @@
 {% load wagtailimages_tags %}
 
-<div class="profile-image {% if modifier %}profile-picture--{{ modifier }}{% endif %}">
+<div class="profile-image {% if modifier %}profile-image--{{ modifier }}{% endif %}">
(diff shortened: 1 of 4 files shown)


rca/project_styleguide/templates/patterns/atoms/profile-image/profile-image.html
```diff
@@ -0,0 +1,6 @@
+{% load wagtailimages_tags %}
+
+<div class="profile-image {% if modifier %}profile-picture--{{ modifier }}{% endif %}">
```
chris-lawton: should the modifier class be `profile-image--{{ modifier }}` rather than `profile-picture--`?

chris-lawton:
Looks wicked! Nice work 👏

I wonder if we should put a max height on the image as it looks like the key details section overlaps the image on wider screens.

Screenshot 2020-03-19 at 09 22 35

rca/project_styleguide/templates/patterns/atoms/profile-image/profile-image.html
nicklee: oops, missed during refactor! Thanks!

PR #335, opened 2020-03-24
Feature/292 access planit data

kevinhowbrook:
`feature/short-courses` is acting as the release branch for master so this is effectively the prod merge.

Tickets:
- Initial Back End https://projects.torchbox.com/projects/rca-website-rebuild/tickets/292
- Booking bar https://projects.torchbox.com/projects/rca-website-rebuild/tickets/296
- Booking modal https://projects.torchbox.com/projects/rca-website-rebuild/tickets/297

### Summary
This PR puts in place the ability to use data from the accessplanit XML feeds on short course pages.
Highlights:
 - XML data is fetched via cron using a management command
 - XML data is parsed down to a digestible list for the template
 - The solution here should expect and work with Timeout errors raised.
 - The parsed data should come from the cache using a key with the course ID
 - Keep the model dumb, it should just call a `get_data()` method and not be concerned with errors

TODOs
- [x] After PR review, review code comments and TODOs
- [x] When ready for stage - set sandbox env vars defined in settings


rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,151 @@
...
+"""
+
+
+# TODO remove this, it's just for debugging
+def print_message(msg):
+    print("\n")
+    print("=" * len(msg))
+    print(f"{msg}")
+    print("=" * len(msg))
+    print("\n")
```
kevinhowbrook: To remove

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,151 @@
...
+        AccessPlanitException: Should the fetch encouter Timeouts or any other
+        exceptions, this raised and caught when setting the data in the cache.
+    """
+
+    def __init__(self, course_id):
+        # This might be better as a default on the model
+        # but if there is no course ID passed we seem to get some default xml
+        # data back, so set it as 0
+        self.course_id = course_id if course_id else "0"
+        self.company_id = "ROYALC9RCH"  # TODO - settings field perhaps?
```
kevinhowbrook: Perhaps make this a settings field if time allows, as it might be different in the live instance of the XML, at the moment we are using a sandbox

rca/shortcourses/models.py
```diff
@@ -54,4 +57,47 @@ class ShortCoursePage(BasePage):
...
+                    context["booking_bar"]["message"] = "Next course starts"
+                    context["booking_bar"]["date"] = access_planit_data[key][
+                        "start_date"
+                    ]
+                    context["booking_bar"][
+                        "action"
+                    ] = f"Book now from \xA3{access_planit_data[key]['cost']}"
+                    context["booking_bar"]["link"] = None
+                    context["booking_bar"]["modal"] = "booking-details"
+                    break
```
kevinhowbrook: This is all a bit too much work being done in the context method, would be better as an internal method on the class

rca/project_styleguide/templates/patterns/atoms/modal-launcher/modal-launcher.html
```diff
@@ -1,4 +1,6 @@
-<button data-micromodal-trigger="{{ modal }}" class="modal-launcher">
+
+<button data-micromodal-trigger="{{ item.modal }}" class="modal-launcher"
+    {% if not item.modal %}onClick="parent.location='{{ item.link }}'" {% endif%}>
```
kevinhowbrook: @nicklee I've assigned you as a reviewer here so you can check you are happy with how this is done.
if there is no `item.modal` coming for the models context we instead use a default link which goes to a form with the course id passed.

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,217 @@
...
+    def test_data_if_timeout(self, mock_get):
+        """ If a timeout is caught the xml_data should be an empty list"""
+        # logging.disable(logging.CRITICAL)  # Don't log exceptions for tests... is this bad?
+        mock_get.side_effect = Timeout
+        data = AccessPlanitXML(course_id=1)
+        xml_data = data.get_data()
+        self.assertEqual(xml_data, [])
+
+    def test_management_command_fetch_data(self):
+        """ Test that xml with no dates goes in the cache as [] """
```
richbrennan: Noting that we discussed setting the cache to `[]` with no dates can be tested outside of the management command, and that testing the command could consist of mocking `AccessPlanitXML` and checking how its used.

It might still be worth having an integration test that deliberately isn't isolated in this way in order to test running the command and the eventual outcome.

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,217 @@
...
+        call_command("fetch_access_planit_data", *args, **opts)
+        for i in range(5):
+            cache_key = f"short_course_{i}"
+            self.assertEqual(cache.get(cache_key), [])
+
+    @mock.patch(
+        "rca.shortcourses.access_planit.AccessPlanitXML.fetch_data_from_xml",
+        side_effect=mocked_fetch_data_from_xml,
+    )
+    def test_management_command_fetch_with_example_data(
```
richbrennan: This is a good integration test for the management command

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,217 @@
...
+        self.assertContains(response, "Short course title")
+        self.assertEqual(
+            response.render().status_code, 200
+        )  # will render 404 if there is an failure
+
+    @mock.patch(
+        "rca.shortcourses.access_planit.AccessPlanitXML.fetch_data_from_xml",
+        side_effect=mocked_fetch_data_from_xml,
+    )
+    def test_page_renders_good_xml(self, mocked_fetch_data_from_xml):
```
richbrennan: Good page integration test 👍 

(Tests that use the Django test client to `.get()` can be viewed as integration tests)

rca/utils/models.py
```diff
@@ -429,3 +430,29 @@ class Meta:
             [FieldPanel("disable_apply_tab")], "Global settings for programme pages"
         )
     ]
+
+
+@register_snippet
+class ShortCourseDetailSnippet(models.Model):
+    SHORT_COURSE_DETAIL_TYPES = [(1, "FAQs"), (2, "T&Cs")]
```
richbrennan: Separating out the choices values into separate constants means you don't have to repeat the value below:

```
FAQ = 1
T&C = 2
SHORT_COURSE_DETAIL_TYPES = [(FAQ, "FAQs"), (T&C, "T&Cs")]

def body(self):
        if self.snippet_type == FAQ:
            ...
```

More info https://www.b-list.org/weblog/2007/nov/02/handle-choices-right-way/

rca/shortcourses/access_planit.py
kevinhowbrook: Thanks, Done

rca/project_styleguide/templates/patterns/molecules/short-course-details/short-course-details.html
```diff
@@ -17,9 +17,9 @@ <h2 class="short-course-details-item__heading heading heading--five">Upcoming da
                     <h2 class="short-course-details-item__heading heading heading--five">Course details</h2>
                 </div>
                 <div class="short-course-details-item__content rich-text">
-                    <p>Early bird offer until 10 March 2020. Booking closes 18 March 2020 or when fully booked.</p>
-                    <p>Please be sure to read our <a href="#">Terms & Conditions</a> before applying.</p>
-                    <p>For more information, please visit our <a href="#">FAQs</a>.</p>
+                    {{ page.course_details_text|safe }}
```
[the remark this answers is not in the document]
kevinhowbrook: Yep thanks, added

rca/project_styleguide/templates/patterns/molecules/short-course-details/short-course-details.html
```diff
@@ -17,9 +17,9 @@ <h2 class="short-course-details-item__heading heading heading--five">Upcoming da
...
+                    {{ page.terms_and_conditions.body }}
+                    {{ page.frequently_asked_questions.body }}
```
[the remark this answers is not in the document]
kevinhowbrook: If added the 'safe' filter for these

rca/settings/base.py
```diff
@@ -704,3 +704,12 @@
 # The API url to pull content from for the homepage, see rca.api_content.content
 API_CONTENT_BASE_URL = env.get("API_CONTENT_BASE_URL", "https://rca.ac.uk")
 DATA_UPLOAD_MAX_NUMBER_FIELDS = 2000
+
+# Access planit cache timeout
+ACCESS_PLANIT_XML_FEED_TIMEOUT = int(env.get("ACCESS_PLANIT_XML_FEED_TIMEOUT", 60 * 60))
```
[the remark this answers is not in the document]
kevinhowbrook: done (i think). The only way I could see how to do this was by trying to cast the value to an int:

```

try:
    ACCESS_PLANIT_XML_FEED_TIMEOUT = int(env.get("ACCESS_PLANIT_XML_FEED_TIMEOUT"))
except TypeError:
    ACCESS_PLANIT_XML_FEED_TIMEOUT = 60 * 60
```

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,151 @@
...
+    in the low-level cache.
+
+    Raises:
+        AccessPlanitException: Should the fetch encouter Timeouts or any other
+        exceptions, this raised and caught when setting the data in the cache.
+    """
+
+    def __init__(self, course_id):
+        # This might be better as a default on the model
+        # but if there is no course ID passed we seem to get some default xml
```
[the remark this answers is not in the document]
kevinhowbrook: The course ID is now a required field (I've added a test that should fail if this changes) https://github.com/torchbox/rca-wagtail-2019/blob/feature/292-access-planit-data/rca/shortcourses/tests/test_access_planit.py#L129

So `self.course_id = course_id if course_id else "0"` is a sort of fail safe if, a) someone makes this field non required and b) the same someone doesn't fix the tests

rca/project_styleguide/templates/patterns/atoms/modal-launcher/modal-launcher.html
nicklee: looks good to me, we'll be able to use that same logic for adding a modifier to tell the modal launcher whether it needs to be the larger version (larger visusally - when it doesn't link to a modal the action text is a lot longer).

I might end up renaming the modal-launcher, as it's obvious now that this doesn't always launch a modal, but this can be done later.

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,151 @@
...
+                    "spaces_available": i.spacesavailable.text,
+                    "status": i.status.text,
+                    "book_now_url": i.booknowurl.text,
+                    "enquire_url": i.enquireurl.text,
+                }
+                items.append(item)
+
+        return items
+
+    def get_parsed_data(self):
```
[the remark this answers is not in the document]
kevinhowbrook: you're right, removed this

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,217 @@
...
+        for i in range(5):
+            ShortCoursePage.objects.create(
+                title=f"Short course {i}",
+                path=str(i),
+                depth="001",
+                access_planit_course_id=i,
+            )
+        args = []
+        opts = {}
+        call_command("fetch_access_planit_data", *args, **opts)
```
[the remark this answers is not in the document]
kevinhowbrook: Nope they don't, I've removed them

rca/utils/models.py
kevinhowbrook: :+1:

rca/shortcourses/tests/test_access_planit.py
kevinhowbrook: Thanks Rich, I've added a test that just instatiates the AccessPlanitXML class with a course id that would cause this and tested the return for the data = [].

rca/shortcourses/tests/test_access_planit.py
kevinhowbrook: :tada:

rca/shortcourses/tests/test_access_planit.py
kevinhowbrook: :tada:

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,151 @@
...
+    The data is periodically fetched with a management command and set
+    in the low-level cache.
+
+    Raises:
+        AccessPlanitException: Should the fetch encouter Timeouts or any other
+        exceptions, this raised and caught when setting the data in the cache.
+    """
+
+    def __init__(self, course_id):
+        # This might be better as a default on the model
```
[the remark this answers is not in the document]
kevinhowbrook: I was thinking it might be better to force the default as 0 at the model level, but I've since made the course ID a required field

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def parse_data(self, xml_data):
+        parser = AccessPlanitXMLParser(xml=xml_data)
+        return parser.get_parsed_data()
+
+    def fetch_data_from_xml(self):
+        url = self.prepare_query()
+        try:
+            response = requests.get(url=url)
+            xml = self.parse_data(response.text)
```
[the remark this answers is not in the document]
kevinhowbrook: Yep, looks like it can, will change that

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,151 @@
...
+                f"Error occurred fetching XML data for course_id: {self.course_id}"
+            )
+            raise AccessPlanitException
+        return xml
+
+    def set_data_in_cache(self):
+        cache_key = f"short_course_{self.course_id}"
+        try:
+            data = self.fetch_data_from_xml()
+            cache.set(cache_key, data, settings.ACCESS_PLANIT_XML_FEED_TIMEOUT)
```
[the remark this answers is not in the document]
kevinhowbrook: Yep, thanks.

rca/shortcourses/models.py
```diff
@@ -54,4 +72,58 @@ class ShortCoursePage(BasePage):
...
+                    booking_bar["link"] = None
+                    booking_bar["modal"] = "booking-details"
+                    break
+        return booking_bar
+
+    def get_context(self, request, *args, **kwargs):
+        context = super().get_context(request, *args, **kwargs)
+        access_planit_data = self.get_access_planit_data()
+        context["register_interest_link"] = register_interest_link = (
+            "https://rca-verdant-staging.herokuapp.com/short-courses/regi"
```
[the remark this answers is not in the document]
kevinhowbrook: Good point, the register form isn't live yet but it needs to point to the legacy site. I'll swap this for an env var.

rca/shortcourses/models.py
```diff
@@ -54,4 +72,58 @@ class ShortCoursePage(BasePage):
...
+        # If there are no dates the booking link should go to a form, not open
+        # a modal, this link is also used as a generic interest link too though.
+        if self.access_planit_course_id:
+            booking_bar["link"] = register_interest_link
+
+        if access_planit_data:
+            for key, date in enumerate(access_planit_data):
+                if date["status"] == "Available":
+                    booking_bar["message"] = "Next course starts"
+                    booking_bar["date"] = access_planit_data[key]["start_date"]
```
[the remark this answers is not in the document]
kevinhowbrook: Yea you are right, because of the `break` I don't need to specify the key. Will amend.

rca/shortcourses/models.py
```diff
@@ -54,4 +72,58 @@ class ShortCoursePage(BasePage):
...
+        }
+        # If there are no dates the booking link should go to a form, not open
+        # a modal, this link is also used as a generic interest link too though.
+        if self.access_planit_course_id:
+            booking_bar["link"] = register_interest_link
+
+        if access_planit_data:
+            for key, date in enumerate(access_planit_data):
+                if date["status"] == "Available":
+                    booking_bar["message"] = "Next course starts"
```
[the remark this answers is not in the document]
kevinhowbrook: Yea this is intended. I only want to change this data `if date["status"] == "Available"`, otherwise the earlier booking_bar definition should be used.

rca/shortcourses/models.py
```diff
@@ -54,4 +72,58 @@ class ShortCoursePage(BasePage):
...
+        this date in the booking bar. If there are no courses available, add
+        a default message."""
+
+        booking_bar = {
+            "message": "Applications are now closed",
+            "action": "Register your interest for upcoming dates",
+        }
+        # If there are no dates the booking link should go to a form, not open
+        # a modal, this link is also used as a generic interest link too though.
+        if self.access_planit_course_id:
```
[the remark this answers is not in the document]
kevinhowbrook: Good point, also will help to break the code if it's ever made optional

rca/shortcourses/models.py
```diff
@@ -54,4 +72,58 @@ class ShortCoursePage(BasePage):
...
+                SnippetChooserPanel("frequently_asked_questions"),
+                SnippetChooserPanel("terms_and_conditions"),
+            ],
+            heading="course details",
+        ),
     ]
+
+    def get_access_planit_data(self):
+        data = AccessPlanitXML(course_id=self.access_planit_course_id)
+        return data.get_data()
```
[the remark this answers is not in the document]
kevinhowbrook: Yea good point, a lot of this is _data_ and that's not really usefule

rca/shortcourses/management/commands/fetch_access_planit_data.py
```diff
@@ -0,0 +1,25 @@
...
+    def handle(self, **options):
+        short_courses = ShortCoursePage.objects.exclude(access_planit_course_id="")
+        for course in short_courses:
+            try:
+                logger.info(
+                    f"Fetching XML for course_id:{course.id} with AccessPlanit ID:{course.access_planit_course_id}"
+                )
+                ap_data = AccessPlanitXML(course_id=course.access_planit_course_id)
+                ap_data.set_data_in_cache()
+            except ValidationError:
```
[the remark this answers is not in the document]
kevinhowbrook: Now I think about it, I don't think I need this in a try/except anymore. I think it was for a course ID that was blank when I wrote it originally, by I'm only running this on courses _with_ course ids now so I'll remove it.

rca/shortcourses/management/commands/fetch_access_planit_data.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+class Command(BaseCommand):
+    help = "Fetches data for each short course page from AccessPlanit XML feed and places it in the cache"
+
+    def handle(self, **options):
+        short_courses = ShortCoursePage.objects.exclude(access_planit_course_id="")
+        for course in short_courses:
+            try:
+                logger.info(
```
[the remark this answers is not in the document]
kevinhowbrook: I'll mode this out of the loop and make the log more generic. It's useful to see when it's running but you are right, it's likely to run every 15 mins so will create too much noise

rca/shortcourses/migrations/0006_add_short_course_details.py
```diff
@@ -0,0 +1,43 @@
+# Generated by Django 2.2.9 on 2020-03-24 16:54
```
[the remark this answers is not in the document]
kevinhowbrook: They could, but because they are on staging at the moment, I want to leave it. It's largely down to this feature being over multiple tickets.

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,261 @@
...
+        self.assertEqual(cache.get(cache_key), self.expected_data)
+        # Add a timeout for the new data request
+        mock_get.side_effect = Timeout
+        AccessPlanitXML(course_id=1).get_data()
+
+        # Now check the stale cache data is there
+        self.assertEqual(cache.get(cache_key), self.expected_data)
+
+    def test_required_course_id(self):
+        """ The course id is a required field, however if this is changed we want some tests to fail,
```
richbrennan, line 156: If what is changed, the field being required or the ID itself?

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,261 @@
...
+        with self.assertRaises(TypeError):
+            ShortCoursePage.objects.create(
+                title=f"Short course should not save",
+                path=100,
+                depth="001",
+                access_planit_course_id="course id cannot be a string",
+            )
+
+    def test_management_command_fetch_data(self):
+        """ Test that xml with no dates doesn't go in the cache """
```
richbrennan, line 172: How is the test condition "xml with no dates" being set up in this test?
Is it doing the same as below, testing a non valid course ID?

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,261 @@
...
+        for i in range(5):
+            ShortCoursePage.objects.create(
+                title=f"Short course {i}",
+                path=str(i),
+                depth="001",
+                access_planit_course_id=i,
+            )
+        args = []
+        opts = {}
+        call_command("fetch_access_planit_data", *args, **opts)
```
richbrennan, line 208: Are `*args, **opts` needed?

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,261 @@
...
+        )
+        home_page.add_child(instance=short_course_page)
+        response = self.client.get("/short-course/")
+        self.assertTemplateUsed(
+            response, "patterns/pages/shortcourses/short_course.html"
+        )
+        self.assertContains(response, "Short course title")
+        self.assertEqual(
+            response.render().status_code, 200
+        )  # will render 404 if there is an failure
```
richbrennan, line 231: Which failure will raise a 404?

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,261 @@
...
+        home_page.add_child(instance=short_course_page)
+        response = self.client.get("/short-course/")
+
+        self.assertTemplateUsed(
+            response, "patterns/pages/shortcourses/short_course.html"
+        )
+        self.assertContains(response, "Short course title")
+        self.assertEqual(
+            response.render().status_code, 200
+        )  # will render 404 if there is an failure
```
richbrennan, line 253: As above, which failure will raise a 404?

rca/shortcourses/tests/test_access_planit.py
kevinhowbrook: I'll remove this - was just my notes when writing the tests

rca/shortcourses/tests/test_access_planit.py
kevinhowbrook: I'll remove this - was just my notes when writing the tests

rca/shortcourses/tests/test_access_planit.py
kevinhowbrook: Nope, removed

rca/shortcourses/tests/test_access_planit.py
kevinhowbrook: I think it _is_ doing the same as below, it's just checking the data return for a course_id passed that won't return anything. The 'XML with no dates' is basically just no data, but I think I'm going to remove this because it similar the below test.

rca/shortcourses/tests/test_access_planit.py
kevinhowbrook: I'll fix this comment, it's a bit vague

rca/settings/base.py
```diff
@@ -700,7 +700,27 @@
...
+    ACCESS_PLANIT_XML_FEED_TIMEOUT = 60 * 60
+
+ACCESS_PLANIT_XML_BASE_URL = env.get(
+    "ACCESS_PLANIT_XML_BASE_URL",
+    "https://rca.accessplanit.com/accessplansandbox/services/WebIntegration.asmx/GetCoursesPackage?",
+)
+ACCESS_PLANIT_SCHOOL_ID = env.get("ACCESS_PLANIT_SCHOOL_ID")
+ACCESS_PLANIT_REGISTER_INTEREST_BASE = env.get(
+    "ACCESS_PLANIT_REGISTER_INTEREST_BASE",
+    "https://rca-verdant-staging.herokuapp.com/short-courses/register-your-interest/",
```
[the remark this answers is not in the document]
kevinhowbrook: Yea I'll change this to `None` and rework... it should always use an env var but we can set those in local.py

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def fetch_data_from_xml(self):
+        url = self.prepare_query()
+        try:
+            response = requests.get(url=url)
+        except Timeout:
+            logger.exception(
+                f"Timeout occurred fetching XML data for course_id: {self.course_id}"
+            )
+            raise AccessPlanitException
```
richbrennan, line 113: Is it worth adding the message to `AccessPlanitException`?

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,139 @@
...
+        except Exception:
+            logger.exception(
+                f"Error occurred fetching XML data for course_id: {self.course_id}"
+            )
+            raise AccessPlanitException
+        else:
+            xml = self.parse_data(response.text)
```
richbrennan, line 120: Could return here

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,139 @@
...
+        return xml
+
+    def set_data_in_cache(self):
+        try:
+            data = self.fetch_data_from_xml()
+        except AccessPlanitException:
+            pass
```
richbrennan, line 127: Might be worth adding a comment why the exception is passed

rca/shortcourses/access_planit.py
```diff
@@ -0,0 +1,139 @@
...
+        except Exception:
+            logger.exception(
+                f"Error occurred fetching XML data for course_id: {self.course_id}"
+            )
+            raise AccessPlanitException
+        else:
+            xml = self.parse_data(response.text)
+        return xml
+
+    def set_data_in_cache(self):
```
richbrennan, line 123: This method is doing more than just setting the data in the cache, its fetching and returning it too. I'd consider renaming it, or possibly breaking it up so its not doing too much, see https://www.martinfowler.com/bliki/CommandQuerySeparation.html

rca/shortcourses/models.py
```diff
@@ -54,4 +76,70 @@ class ShortCoursePage(BasePage):
...
+                    booking_bar["action"] = f"Book now from \xA3{date['cost']}"
+                    booking_bar["link"] = None
+                    booking_bar["modal"] = "booking-details"
+                    break
+        return booking_bar
+
+    def clean(self):
+        errors = defaultdict(list)
+        try:
+            int(self.access_planit_course_id)
```
richbrennan, line 124: If `access_planit_course_id` has to be an `int`, could the field be made an integer?

rca/shortcourses/tests/test_access_planit.py
```diff
@@ -0,0 +1,248 @@
...
+        self.assertEqual(xml_data, None)
+
+    @mock.patch(
+        "rca.shortcourses.access_planit.AccessPlanitXML.fetch_data_from_xml",
+        side_effect=mocked_fetch_data_from_xml,
+    )
+    @mock.patch("rca.shortcourses.access_planit.requests.get")
+    def test_cached_data_if_timeout(self, mock_get, mocked_fetch_data_from_xml):
+        """ Test getting stale cache data with a Timeout failure to get data"""
+        # logging.disable(logging.CRITICAL)
```
richbrennan, line 146: Should this be commented?

rca/shortcourses/access_planit.py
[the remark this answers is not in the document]
kevinhowbrook: Looking at this again I can probably remove this 0... seen as though it's a required field. You're right it is a bit harder to follow

rca/shortcourses/tests/test_access_planit.py
kevinhowbrook: nope, tah

rca/shortcourses/access_planit.py
kevinhowbrook: Yep, will do, thanks

rca/shortcourses/models.py
kevinhowbrook: done this, thanks

richbrennan: see caveat above about whether course ids could start with zero

richbrennan: https://github.com/torchbox/rca-wagtail-2019/pull/335#pullrequestreview-393006124

rca/shortcourses/access_planit.py
kevinhowbrook: I'm out of time to make a change like this, I'll change the name of the method to something a bit clearer as to what it's doing

PR #350, opened 2020-04-06
Feature/guide fe

nicklee:


rca/static_src/sass/components/_card.scss
```diff
@@ -459,6 +459,119 @@
...
+                display: block;
+                height: $notch-height-small;
+                width: 50%;
+
+                @include media-query(medium) {
+                    height: $notch-height-medium;
+                }
+
+                @include media-query(large) {
+                    content: none;
```
chris-lawton, line 509: I never thought of doing this to hide the content, always used `display: none`

nicklee: I'm never 100% sure which is best, content: none always feels less heavy to me, but can't say that has any basis!

PR #358, opened 2020-04-08
Implement base models for research section

kevinhowbrook:
[Ticket 315] https://projects.torchbox.com/projects/rca-website-rebuild/tickets/315

Adds base models for the research section


rca/research/models.py
```diff
@@ -0,0 +1,6 @@
+from rca.utils.models import BasePage
+
+
+class ResearchCentrePage(BasePage):
+    template = "patterns/pages/researchcentre/research_centre.html"
+    pass
```
kevinhowbrook, line 6: Technically this doesn't need to be here... it's about to change in another pr though

rca/landingpages/migrations/0002_rename_model.py
```diff
@@ -0,0 +1,24 @@
+# Generated by Django 2.2.9 on 2020-04-14 09:53
```
richbrennan, line 1: Could the migrations be consolidated easily? (not break anyone elses build if deleted and recreated as one)

kevinhowbrook: I wanted to do this but it's already gone to staging so I'm leaving it as is

PR #360, opened 2020-04-14
Feature/374 research back end

kevinhowbrook:
Tickets
- https://projects.torchbox.com/projects/rca-website-rebuild/tickets/377
- https://projects.torchbox.com/projects/rca-website-rebuild/tickets/375
- https://projects.torchbox.com/projects/rca-website-rebuild/tickets/374


richbrennan:
Some if this looks familiar, is it waiting on another MR?

richbrennan:
Also there are some conflicts to resolve

kevinhowbrook:
> Some if this looks familiar, is it waiting on another MR?

@richbrennan this is ready now - I've rebased it on `feature/phase-2` which will be our release branch.

nicklee:
Yaml updated 👍

rca/research/models.py
```diff
@@ -1,6 +1,313 @@
...
+    )
+    primary_staff_name = models.CharField(
+        blank=True, max_length=250, help_text=_("The name of the staff member")
+    )
+    primary_staff_role = models.CharField(
+        blank=True,
+        max_length=120,
+        help_text=_("The role of the staff member, E.G 'Head of Programme'"),
+    )
+    centre_address = RichTextField(blank=True)
```
richbrennan, line 100: Might it be worth limiting the rich text features?

rca/research/models.py
```diff
@@ -1,6 +1,313 @@
...
+            ObjectList(content_panels, heading="Content"),
+            ObjectList(key_details_panels, heading="Key details"),
+            ObjectList(BasePage.promote_panels, heading="Promote"),
+            ObjectList(BasePage.settings_panels, heading="Settings"),
+        ]
+    )
+
+    def get_child_projects(self):
+        """
+        Returns a queryset of all child ProjectPages of this page
```
richbrennan, line 202: This returns a list

rca/research/models.py
```diff
@@ -1,6 +1,313 @@
...
+                    "image": page.listing_image,
+                    "description": page.listing_summary,
+                }
+            )
+        return child_projects
+
+    def get_research_spaces(self):
+        research_spaces = []
+        for value in self.research_spaces.select_related("page"):
+            if value.page.live:
```
richbrennan, line 228: The `RelatedPage.page` field can be `null`, which would cause an `AttributeError` here if it were. https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/utils/models.py#L98

I don't think `RelatedPage.page` should be nullable. Its `on_delete=models.CASCADE` so a null value is less likely, but might be worth changing to be safe.

rca/research/models.py
```diff
@@ -1,6 +1,313 @@
...
+            errors["staff_link_text"].append(
+                "Please add some text for the link to all staff"
+            )
+
+        if errors:
+            raise ValidationError(errors)
+
+    def get_context(self, request, *args, **kwargs):
+        context = super().get_context(request, *args, **kwargs)
+        context["hero_colour"] = "dark"
```
richbrennan, line 299: Ideally `dark` and `light` would constants, especially as they're used here too https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/home/models.py#L208

rca/research/models.py
```diff
@@ -1,6 +1,313 @@
...
+        if int(self.hero_colour_option) == LIGHT_TEXT_ON_DARK_IMAGE:
+            context["hero_colour"] = "light"
+
+        context["about_page"] = self.about_page_url
+        if self.about_page:
+            context["about_page"] = self.about_page.url
```
richbrennan, line 305: These three lines could be:

`context["about_page"] = self.about_page.url if self.about_page else self.about_page_url`

richbrennan:
A few comments, thanks Kevin!

rca/research/models.py
kevinhowbrook: I understand this, If I remove the related page here, everything works fine. Are you suggesting that the I change the RelatedPage.page so it's not nullable? That sound like a safe move, my concern is handling the default values for existing entries, would it just be a blank string because blank=True?

richbrennan: You'd also remove `blank=True` (so it'd use the default `blank=False`).

Yes ideally `RelatedPage.page` would be not nullable, but you're right in that default values for existing records could be an issue as it'll want a valid page id.

You could maybe have `if value.page and value.page.live:` instead

kevinhowbrook: Yea good shout - I'd still like to change that related page null though, seems to scare me having it hanging around. I'll add in a check for the page object too and create a follow issue for this https://github.com/torchbox/rca-wagtail-2019/issues/378

PR #362, opened 2020-04-15
Update vagrant Postgres

Stormheg:
While provisioning the vagrant vm I hit an error with Postgres not starting. This can be fixed by explicitly starting Postgres.
I've also updated Postgres to version 12 to avoid version mismatches.


kevinhowbrook:
Thanks @Stormheg

PR #363, opened 2020-04-15
Add workaround for missing linux headers in vagrant

Stormheg:
This should fix an error that occurs due to missing linux kernel headers when you provision the vagrant vm.


kevinhowbrook:
cheers @Stormheg

PR #373, opened 2020-04-22
Swap for ReadTimeout because it extends Timeout and is being caught b…

kevinhowbrook:
…y sentry

see [sentry issue](https://sentry.io/organizations/torchbox/issues/1603072847/?project=1542908&query=is%3Aunresolved)

Previous task was to catch the Timeout's being thrown if the api request took too long (10 senconds). However sentry is catching [ReadTimeout](https://2.python-requests.org/en/v3.0.0/_modules/requests/exceptions/#ReadTimeout) (making the logs noisy)

### DIFF
--- a/rca/api_content/content.py
+++ b/rca/api_content/content.py
@@ -5,7 +5,7 @@ import requests
 from bs4 import BeautifulSoup
 from django.conf import settings
 from django.http.request import QueryDict
-from requests.exceptions import ConnectionError, HTTPError, Timeout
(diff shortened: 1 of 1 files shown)


richbrennan:
According to the[ requests docs](https://requests.readthedocs.io/en/master/api/#requests.Timeout) catching `Timeout` should also catch `ReadTimeout `. 

I wonder if this is related to the [ticket](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/417) for the `HTTPError` that doesn't seem to be caught either?

PR #374, opened 2020-04-23
Add breadcrumbs

kevinhowbrook:


nicklee:
Have fixed all of the above other than "project_listing" as this template isn't in this branch yet

nicklee:
@kevinhowbrook think this is good to be merged in now

PR #376, opened 2020-04-24
Feature/project picker page

kevinhowbrook:
Depends on https://github.com/torchbox/rca-wagtail-2019/pull/372

Logic for the filters (singular of multiple) is going to be discussed on 28/4/2020 so best to review after that. At the moment I'm assuming multiple filters should be possible, but we may be switching this to be a lot simpler; the filters would become single links and the list would only be filterable by 1 of the any items

Logic and implementation of this needs discussion, in particular:
- Pagination and query parameter implementation
- Queryset filtering


rca/projects/models.py
```diff
@@ -265,4 +273,164 @@ def get_context(self, request, *args, **kwargs):
...
+
+        if school_or_centre:
+            projects = projects.filter(
+                models.Q(related_school_pages__page_id__in=school_or_centre)
+                | models.Q(related_research_pages__page_id__in=school_or_centre)
+            )
+        return self._format_results(projects.distinct())
+
+    def get_context(self, request, *args, **kwargs):
+        page = request.GET.get("page", 1)
```
kevinhowbrook, line 409: This will be changed to a settings var, kept it like this for testing

rca/project_styleguide/templates/patterns/pages/project/project_listing.html
```diff
@@ -52,32 +51,45 @@ <h1 class="heading heading--display-two">{{ page.title }}</h1>
...
                     </a>
                 </div>
-
             </section>
 
             <section class="section bg bg--light">
+ <form action="" method="get">
+                    {% for item in filters %}
+                        <ul>
+                            {{item.title}}
```
richbrennan, line 60: Is having `{{item.title}}` between `<ul>` and `<li>` valid markup?

rca/projects/models.py
```diff
@@ -237,6 +240,11 @@ def clean(self):
         if errors:
             raise ValidationError(errors)
 
+    def get_related_school(self):
+        """ returns the first realated schools page"""
```
richbrennan, line 244: Typo

rca/projects/models.py
```diff
@@ -237,6 +240,11 @@ def clean(self):
...
+        if self.related_school_pages.first():
+            return self.related_school_pages.first().page
```
richbrennan, line 246: Maybe avoid running the query twice

rca/projects/models.py
```diff
@@ -265,4 +273,164 @@ def get_context(self, request, *args, **kwargs):
...
+                    "id": i.id,
+                    "title": i.title,
+                    "link": slugify(i.title),
+                    "filter_type": "subject",
+                }
+            )
+        filters.append(subjects)
+
+        school_or_centre = {"title": "School or centre", "items": []}
+        for i in SchoolPage.objects.all():
```
richbrennan, line 324: Is a live/public filter needed?

rca/projects/models.py
```diff
@@ -265,4 +273,164 @@ def get_context(self, request, *args, **kwargs):
...
+            school_or_centre["items"].append(
+                {
+                    "id": i.id,
+                    "title": i.title,
+                    "link": slugify(i.title),
+                    "filter_type": "school_or_centre",
+                }
+            )
+        filters.append(school_or_centre)
+        for i in ResearchCentrePage.objects.all():
```
richbrennan, line 334: Is a live/public filter needed?

rca/projects/models.py
```diff
@@ -265,4 +273,164 @@ def get_context(self, request, *args, **kwargs):
...
+        project_results = self.get_results(request, context)
+        context["results_count"] = len(project_results)
+
+        # Send all the query params through to the context so they can be added
+        # to the pager links, E.G type=1&type=2&subject=1...
+        context["extra_query_params"] = self.get_extra_query_params(request)
+
+        # Don't show the featured project if queries are being mage
+        context["show_featured_project"] = True
+        if self.get_extra_query_params(request):
```
richbrennan, line 423: Could this be `if context["extra_query_params"]:` ?

rca/projects/models.py
```diff
@@ -265,4 +273,164 @@ def get_context(self, request, *args, **kwargs):
...
+        return self._format_results(projects.distinct())
+
+    def get_context(self, request, *args, **kwargs):
+        page = request.GET.get("page", 1)
+        context = super().get_context(request, *args, **kwargs)
+        context["filters"] = self.get_filters()
+        context["featured_project"] = self.featured_project
+
+        project_results = self.get_results(request, context)
+        context["results_count"] = len(project_results)
```
richbrennan, line 415: The paginator objects provides a `.count` attribute, could that be used instead?
https://docs.djangoproject.com/en/3.0/topics/pagination/#example

rca/projects/models.py
jafacakes2011: @kevinhowbrook could you explain what will become a setting here?

rca/utils/templatetags/util_tags.py
```diff
@@ -63,3 +63,71 @@ def social_media_links(context):
...
+    When current_page is 9
+        [[1], [8, 9, 10]]
+    When current_page is 10
+        [[1], [8, 9, 10]]
+    """
+    # If there are 4 or less items, just return a list containing 1 section.
+    if len(page_range) <= 4:
+        return [page_range]
+
+    # Build a list of lists for the differnet pagination sections
```
kevinhowbrook, line 92: typo differnet

rca/projects/models.py
kevinhowbrook: https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/settings/base.py#L672

kevinhowbrook: (the default count of items... which I think should be 25)

jafacakes2011: Ahh, the confusion was caused by the line this comment is referring to, it should be the line: `paginator = Paginator(project_results, 24)`

I will fix this now so that we use `DEFAULT_PER_PAGE` which should be 24

rca/projects/models.py
```diff
@@ -336,4 +345,225 @@ def get_context(self, request, *args, **kwargs):
...
+
+        # used_filter_values will equal a dictionary with the same keys as
+        # project_filter_field_mapping and the values of each item will be
+        # a list of IDs.
+        # Example: {
+        #     'type': [1,2],
+        #     'subject': [4,8],
+        #     'school': [],
+        #     'centre': [11,17],
+        # }
```
kevinhowbrook, line 380: I think school and centre are combined right? `school_or_centre`

rca/projects/models.py
```diff
@@ -19,7 +22,7 @@
 from wagtail.images import get_image_model_string
 from wagtail.images.edit_handlers import ImageChooserPanel
 
-from rca.home.models import HERO_COLOUR_CHOICES, LIGHT_TEXT_ON_DARK_IMAGE
+from rca.home.models import HERO_COLOUR_CHOICES, LIGHT_HERO, LIGHT_TEXT_ON_DARK_IMAGE
```
kevinhowbrook, line 25: I've added a separate issue for what we spoke about here https://github.com/torchbox/rca-wagtail-2019/issues/395

rca/projects/models.py
kevinhowbrook: duh, yea my bad. Thanks

rca/projects/models.py
```diff
@@ -336,4 +345,227 @@ def get_context(self, request, *args, **kwargs):
...
+        # to the pager links, E.G type=1&type=2&subject=1...
+        context["extra_query_params"] = self.get_extra_query_params(
+            request, active_filters
+        )
+
+        # Unfiltered Projects query.
+        projects_query = self.get_projects_query()
+
+        context["filters"] = self.get_filters(active_filters, projects_query)
+        context["featured_project"] = self._format_results([self.featured_project])[0]
```
kevinhowbrook, line 545: I think this need wrapping in an if statement as if there are no featured projects selected it will break

PR #379, opened 2020-04-27
Feature/372 projects related projects

kevinhowbrook:
[Ticket 372](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/372)

This is on staging for QA, here is an example (bottom of this page https://rca-staging.herokuapp.com/helen-hamlyn-centre-design/example-project/)

Projects have a 'Key details' tab... in this tab various taxonomies are selected. The projects that show in the 'related projects' area follow the logic defined in the designs (without the theme, because we have not defined  theme yet). E.G

        IF there are no projects with the same theme School/Centre latest projects with a
        matching research_type will be displayed.
        IF there are no projects with a matching research_type, the latest projects with
        matching subject tags will be displayed.


rca/projects/models.py
```diff
@@ -218,6 +218,74 @@ class ProjectPage(BasePage):
...
+        structure for the gallery template
+
+        Arguments:
+            projects: Queryset of project pages to format
+
+        Returns:
+            List: Maximum of 8 projects
+        """
+        items = []
+        for page in projects:
```
richbrennan, line 232: Could this be `for page in projects[:8]:` so not all of the `projects` are processed if not needed?

rca/projects/models.py
```diff
@@ -218,6 +218,74 @@ class ProjectPage(BasePage):
...
+            page = page.specific
+            meta = None
+            if page.related_school_pages.first():
+                meta = page.related_school_pages.first().page.title
```
richbrennan, line 236: Might be best not to run `page.related_school_pages.first()` twice, maybe:
```
meta = ''
related_school = page.related_school_pages.first()
if related_school is not None:
    meta = related_school.page.title
```

rca/projects/models.py
```diff
@@ -218,6 +218,74 @@ class ProjectPage(BasePage):
...
+        IF there are no projects with the same theme School/Centre latest projects with a
+        matching research_type will be displayed.
+        IF there are no projects with a matching research_type, the latest projects with
+        matching subject tags will be displayed.
+
+        Returns:
+            List -- of filtered and formatted ProjectPages
+        """
+
+        all_projects = ProjectPage.objects.live().public().exclude(pk=self.pk)
```
richbrennan, line 260: You could replace `.exclude(pk=self.pk)` with `not_page(self)`

PR #380, opened 2020-04-27
Feature/404 landing page default

kevinhowbrook:
This is the initial PR for the first (of 4) landing pages.
The Landing Page - Default will be a generic landing page, elements of which are in all the other landing pages (along with extras). 

@nicklee assigned you to review the yaml updates here with my changes


rca/landingpages/models.py
```diff
@@ -1,17 +1,242 @@
...
+            "The brief paragraph of text to be displayed above the related pages grid"
+        ),
+    )
+    page_list_title = models.TextField(
+        max_length=80,
+        blank=True,
+        help_text=_("The title to be displayed above the page list blocks"),
+    )
+    page_list = StreamField([("page_list", RelatedPageListBlock())], blank=True)
+
```
kevinhowbrook, line 113: Instead of a related model, I've used streamfields for this. The design shows too many variations

kevinhowbrook: image

rca/landingpages/models.py
```diff
@@ -1,17 +1,242 @@
...
+            for page in block.value["page"]:
+                page = page.value.specific
+                item["related_items"].append(page)
+            items.append(item)
+        return items
+
+    def get_context(self, request, *args, **kwargs):
+        context = super().get_context(request, *args, **kwargs)
+        context["hero_colour"] = DARK_HERO
+        if int(self.hero_colour_option) == LIGHT_TEXT_ON_DARK_IMAGE:
```
richbrennan, line 208: Is the `int()` cast needed?

rca/landingpages/models.py
kevinhowbrook: Good point, but this is a confident no - Pages and Landing pages wont have any actual relationship relationships

PR #387, opened 2020-04-29
refactor alumni stories and news/events to use a management command t…

richbrennan:
…o fetch from api

Adds a management command and related code to fetch content from the old site api outside of the page request proces.

### DIFF
--- a/rca/api_content/content.py
+++ b/rca/api_content/content.py
@@ -278,7 +278,6 @@ def pull_alumni_stories(programme_type_slug=None):
 
 
 class _BaseContentAPI:
-
     def __init__(self, func, cache_key):
         self.func = func
         self.cache_key = cache_key
@@ -298,15 +297,12 @@ class _BaseContentAPI:
         return []
 
 
-
 class AlumniStoriesAPI(_BaseContentAPI):
-
     def __init__(self):
         super().__init__(pull_alumni_stories, "latest_alumni_stories")
 
 
 class NewsEventsAPI(_BaseContentAPI):
-
     def __init__(self):
         super().__init__(pull_news_and_events, "news_and_events_data")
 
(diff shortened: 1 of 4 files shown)


rca/api_content/content.py
```diff
@@ -36,32 +41,28 @@ def ranged_date_format(date, date_to):
...
+            f"Error occured when fetching further detail data from {url}"
         )
     except (HTTPError, ConnectionError):
         logger.exception(
-            "HTTP/ConnectionError occured when fetching further detail data from {url}"
+            f"HTTP/ConnectionError occured when fetching further detail data from {url}"
         )
         raise CantPullFromRcaApi(
-            "Error occured when fetching further detail data from {url}"
+            f"Error occured when fetching further detail data from {url}"
```
kevinhowbrook, line 58: as if I forgot the f string :facepalm:

rca/api_content/content.py
```diff
@@ -274,3 +275,41 @@ def pull_alumni_stories(programme_type_slug=None):
...
+
+
+class _BaseContentAPI:
+    def __init__(self, func, cache_key):
+        self.func = func
+        self.cache_key = cache_key
+
+    def fetch_from_api(self):
+        try:
+            data = self.func()
```
kevinhowbrook, line 287: I've not seen this pattern before but at instantiation wil func = pull_alumni_stories/pull_news_and_events ?

rca/api_content/content.py
richbrennan: I missed it in code review 😬

rca/api_content/content.py
richbrennan: Yes, functions are first class objects so they can be assigned/passed around etc, then called when needed.

richbrennan: TBH this class inheritance might not be the best pattern, but it separated the code into command and query so they can be used in different contexts.

Passing functions around and calling them dynamically is fine though.

PR #388, opened 2020-04-30
If a title value isn't added, we should get the title from the page v…

kevinhowbrook:
…alue

### DIFF
--- a/rca/navigation/models.py
+++ b/rca/navigation/models.py
@@ -83,6 +83,8 @@ class LinkBlock(blocks.StructBlock):
 
         if value["page"] and not value["url"]:
             value["url"] = Page.objects.get(id=value["page"]).url
+            if not value["title"]:
+                value["title"] = Page.objects.get(id=value["page"]).title
 
         return value
 
--- a/rca/navigation/models.py
+++ b/rca/navigation/models.py
@@ -2,7 +2,7 @@ from urllib.parse import urljoin, urlparse, urlsplit, urlunsplit
 
 from django import forms
 from django.core import validators
-from django.core.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


rca/navigation/models.py
```diff
@@ -83,6 +83,8 @@ def get_api_representation(self, value, context=None):
 
         if value["page"] and not value["url"]:
             value["url"] = Page.objects.get(id=value["page"]).url
+            if not value["title"]:
+                value["title"] = Page.objects.get(id=value["page"]).title
```
richbrennan, line 87: If this code path is hit then the page query is run again, so might be best to store the result of the previous query in a variable.
Could it ever be the case that the page referred to in `value["page"]` will have been removed but this reference not updated? If so the `.get()` would raise an exception that could break the page

kevinhowbrook: I've added a catch for this... but it's a bit moot, as the resulting data is cached on the legacy site so stale page IDs could well end up in there anyway.

rca/navigation/models.py
```diff
@@ -82,7 +82,16 @@ def get_api_representation(self, value, context=None):
         )
 
         if value["page"] and not value["url"]:
-            value["url"] = Page.objects.get(id=value["page"]).url
+            # Stale cache data could store a deleted page ID, so try the get
+            # query first
+            try:
+                page = Page.objects.get(id=6)
```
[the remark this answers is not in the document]
kevinhowbrook: Certainly not, fixed in https://github.com/torchbox/rca-wagtail-2019/pull/388/commits/db0322057ef3e82ee2d46a3ca934859e8c86ccd6

PR #389, opened 2020-04-30
Provide a template filter for checking internal or external links

kevinhowbrook:
Example:
```
{% is_external value.href value.url as is_external %}
<a href="{% firstof value.href value.url %}" {% if is_external %} target="_blank" {% endif %}  class="link link--tertiary link--link">````

### CODE REVIEW

File: rca/utils/templatetags/util_tags.py [hunk 1]
```diff
@@ -63,3 +66,46 @@ def social_media_links(context):
...
+    except StopIteration:
+        # incase we get passed empty strings
+        return False
+
+    # Pattern library links are all '#' which will be internal.
+    if link != "#":
+        if urlparse(link).hostname not in default_domains:
+            return True
+
+    return is_external
```
jafacakes2011, line 111: I feel like the `is_external` variable isn't really used in this method. Perhaps we remove it and change this to `return False` or we change line 104 to `return is_external` and line 109 to `is_external = True`

PR #396, opened 2020-05-04
423 - Circle CI changes to the new deployment config

jafacakes2011:
CI Updates to use TBX's new deployment setup

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,7 +69,8 @@ jobs:
       - run:
           name: deploy dev to heroku
           command: |
-            git push https://heroku:$HEROKU_TOKEN@git.heroku.com/$HEROKU_APP_NAME_DEV.git master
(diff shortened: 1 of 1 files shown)


tomusher:
Thanks for sorting this out!

I don't think $CI_COMMIT_SHA is available in CircleCI - might need to replace that with CIRCLE_SHA1.

The DEPLOYMENT_KEY is visible from the Heroku app and should be added as an enironment variable in Circle CI.

I've updated the UPSTREAM_REPO to this Github repo, and added the deploy tool's public key as a read-only deploy key on this repo.

peteandrew:
I've added the DEPLOYMENT_KEY environment variable to CircleCI

cjknowles:
>  and added the deploy tool's public key as a read-only deploy key on this repo.

We couldn't do this, as a public key can only be used as a deployment key in one repo, so we've added a machine user (sysadmin+deploy@torchbox.com) to Github and this repo with read-only access.

PR #400, opened 2020-05-05
Stop duplicating the hero code, it can just come from BasePage context

kevinhowbrook:


rca/utils/models.py
```diff
@@ -351,6 +361,13 @@ class Meta:
         Page.promote_panels + SocialFields.promote_panels + ListingFields.promote_panels
     )
 
+    def get_context(self, request, *args, **kwargs):
+        context = super().get_context(request, *args, **kwargs)
+        context["hero_colour"] = LIGHT_HERO
+        if int(self.hero_colour_option) == DARK_TEXT_ON_LIGHT_IMAGE:
```
jafacakes2011, line 367: Since `hero_colour_option` isn't field on base page, this will raise errors on pages where hero_colour_option isn't defined. 2 options:
 - Move the field up to BasePage
 - Add a conditional to check if self has the `hero_colour_option` attribute.

kevinhowbrook: Thanks - got this in

PR #401, opened 2020-05-06
Fix yaml for guide page

kevinhowbrook:


rca/project_styleguide/templates/patterns/pages/guide/guide.yaml
```diff
@@ -17,11 +17,11 @@ context:
       - dummy
     last_published_at: !!timestamp '2020-05-13 2:00:00'
   anchor_nav:
-    - { title: Introduction, link: '#introduction' }
-    - { title: Article 1, link: '#article-one' }
-    - { title: Article 2, link: '#article-two' }
+    - { title: Introduction, link: '#heading-text' }
+    - { title: Article 1, link: '#heading-text' }
+    - { title: Article 2, link: '#heading-text' }
     - { title: Staff, link: '#staff' }
```
kevinhowbrook, line 23: Only the first dummy block here will work - as they all output an ID of 'heading-text'

PR #409, opened 2020-05-12
286 Add short courses to the programmes listing API and results views

thibaudcolas:
See [tickets/286](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/286) for the initial brief and testing steps. This adds instances of `shortcourses.ShortCoursePage` to the programmes’ listing view, which was so far only displaying instances of `programmes.ProgrammePage`.

This is implemented by doing two separate API calls in parallel, to reduce customisations on the Wagtail API side, after discussion with Kevin and [on Slack](https://torchbox.slack.com/archives/C03R0FNK7/p1589289353027100).

---

I slightly refactored the API field definitions so it’s a bit clearer they are meant to match (not a dealbreaker if they stop matching in the future, as long as it’s intentional, but for now they’re meant to match). The API client makes the two queries in parallel, flattens the results as a single array (that determines the order of the two groups currently, see comment on the ticket).

On the display side, I had to make changes for the programme-only fields to be optional. Decided not to rename too many of the things named "Programme<thing>", but at least update the comments to make it clear there are two page types – and added tests for `ProgrammeTeaser` since it’s where being aware of the two types is the most important.

---

If I had more time I’d probably have a look at API error handling. I can see the error correctly propagates to Redux, but it looks like I had never added something to actually display it ("please refresh the page and try again"). I’ll add a ticket for that.

### DIFF
--- a/rca/programmes/models.py
+++ b/rca/programmes/models.py
@@ -8,6 +8,7 @@ from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.db import models
 from django.utils.text import slugify
 from modelcluster.fields import ParentalKey
+from rest_framework.fields import CharField as CharFieldSerializer
 from wagtail.admin.edit_handlers import (
     FieldPanel,
     InlinePanel,
@@ -596,16 +597,21 @@ class ProgrammePage(BasePage):
         ),
     ]
     api_fields = [
-        APIField("degree_level", serializer=degree_level_serializer()),
+        # Fields for filtering and display, shared with shortcourses.ShortCoursePage.
         APIField("subjects"),
         APIField("programme_type"),
-        APIField("programme_description_subtitle"),
-        APIField("pathway_blocks"),
+        APIField("related_schools_and_research_pages"),
+        APIField(
+            "summary",
+            serializer=CharFieldSerializer(source="programme_description_subtitle"),
+        ),
         APIField(
             name="hero_image_square",
             serializer=ImageRenditionField("fill-580x580", source="hero_image"),
         ),
-        APIField("related_schools_and_research_pages"),
+        # Displayed fields, specific to programmes.
+        APIField("degree_level", serializer=degree_level_serializer()),
+        APIField("pathway_blocks"),
     ]
 
     def get_alumni_stories(self, programme_type_legacy_slug):
--- a/rca/static_src/javascript/programmes/components/ProgrammesResults/ProgrammeTeaser.js
+++ b/rca/static_src/javascript/programmes/components/ProgrammesResults/ProgrammeTeaser.js
@@ -6,8 +6,6 @@ import {
     programmePageShape,
 } from '../../programmes.types';
 
-const shortCourseLabel = 'Short course';
-
 /**
  * A programme or short course’s teaser info, to be displayed as part of search results.
  * "Degree level" and "Pathway blocks" are for programmes only. Other fields are shared.
@@ -15,6 +13,8 @@ const shortCourseLabel = 'Short course';
 const ProgrammeTeaser = ({ programme, onMouseOver, onFocus }) => {
     const { meta, title, summary, degree_level, pathway_blocks } = programme;
     const isShortCourse = meta.type === SHORT_COURSE_PAGE_TYPE;
+    // Short courses have no degree level. Display a distinguishing label instead.
+    const degreeLabel = isShortCourse ? 'Short course' : degree_level.title;
 
     return (
         <a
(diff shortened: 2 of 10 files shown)


rca/programmes/models.py
```diff
@@ -596,16 +597,21 @@ class ProgrammePage(BasePage):
...
-        APIField("degree_level", serializer=degree_level_serializer()),
+        # Fields for filtering and display, shared with shortcourses.ShortCoursePage.
         APIField("subjects"),
         APIField("programme_type"),
-        APIField("programme_description_subtitle"),
-        APIField("pathway_blocks"),
+        APIField("related_schools_and_research_pages"),
+        APIField(
+            "summary",
+            serializer=CharFieldSerializer(source="programme_description_subtitle"),
```
thibaudcolas, line 606: @kevinhowbrook FYI – renaming this field could be problematic if there is other code using the API (I’ve assumed there wasn’t).

rca/static_src/javascript/programmes/components/ProgrammesResults/ProgrammeTeaser.js
```diff
@@ -1,19 +1,20 @@
...
-        meta,
-        title,
-        degree_level,
-        programme_description_subtitle,
-        pathway_blocks,
-    } = programme;
+    const { meta, title, summary, degree_level, pathway_blocks } = programme;
+    const isShortCourse = meta.type === SHORT_COURSE_PAGE_TYPE;
+    // Short courses have no degree level. Display a distinguishing label instead.
+    const degreeLabel = isShortCourse ? 'Short course' : degree_level.title;
```
thibaudcolas, line 17: Note this is the label that might need changing based on feedback on the ticket.

rca/static_src/javascript/programmes/programmes.api.js
```diff
@@ -1,17 +1,33 @@
...
+            'hero_image_square',
+            'degree_level',
+            'pathway_blocks',
+        ],
+    },
+    {
+        type: SHORT_COURSE_PAGE_TYPE,
+        fields: ['summary', 'hero_image_square'],
+    },
+];
```
thibaudcolas, line 29: Note this is the array that might need reordering depending on feedback on the ticket.

We may also need to implement client-side sorting if we want the two types to be mixed – that would need be implemented right after `const items = results.reduce((all, res) => all.concat(res.items), []);`, being careful to only do this for **filtering** responses rather than search results (which should be ordered by relevance server-side).

rca/static_src/javascript/programmes/programmes.types.js
```diff
@@ -1,24 +1,35 @@
...
- * A Programme page’s serialised attributes, coming from the Wagtail API.
- * Please make sure this stays up-to-date with APIField and serializers.
+ * A ProgrammePage or ShortCoursePage’s serialised attributes, coming from the Wagtail API.
+ * Please make sure this stays up-to-date with api_field definitions and serializers.
  */
 export const programmePage = {
     id: PropTypes.number.isRequired,
     meta: PropTypes.shape({
-        type: PropTypes.string.isRequired,
+        type: PropTypes.oneOf([PROGRAMME_PAGE_TYPE, SHORT_COURSE_PAGE_TYPE])
```
thibaudcolas, line 13: Feels a bit overkill but at least it makes it clear this is now for two different page types.

rca/programmes/models.py
kevinhowbrook: Noted, thanks. At the moment there isn't but I also don't expect this to change it. I imagine if we need to we can expose a custom field to the API though to dodge any issues

rca/static_src/javascript/programmes/programmes.api.js
```diff
@@ -42,40 +58,40 @@ export const getWagtailAPIQueryString = ({
...
+            limit: 50,
         });
+        const url = `${WAGTAIL_API_ENDPOINT}${queryString}`;
+
+        return fetch(url, {
+            signal: abortGetProgrammes.signal,
+        }).then((response) => response.json());
+    });
+
+    return Promise.all(requests).then((results) => {
```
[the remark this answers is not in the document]
thibaudcolas: Yeah, the current implementation really sucks and it’s probably my bad for forgetting to add this message initially. I’ve added a ticket for this at https://projects.torchbox.com/projects/rca-website-rebuild/tickets/437.

If I revisit the ordering here I might add it earlier.

rca/static_src/javascript/programmes/components/ProgrammesResults/ProgrammeTeaser.js
```diff
@@ -1,19 +1,20 @@
...
+    SHORT_COURSE_PAGE_TYPE,
+    programmePageShape,
+} from '../../programmes.types';
 
 /**
- * A programme’s teaser info, to be displayed as part of search results.
+ * A programme or short course’s teaser info, to be displayed as part of search results.
+ * "Degree level" and "Pathway blocks" are for programmes only. Other fields are shared.
  */
 const ProgrammeTeaser = ({ programme, onMouseOver, onFocus }) => {
```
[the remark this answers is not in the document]
thibaudcolas: @kevinhowbrook do you have thoughts on whether we’re likely to have another variant of this short to mid term?

PR #427, opened 2020-06-03
[456] Upgrade Wagtail to 2.9

ababic:
This should allow us to utlise the custom tag model functionality that was added in this version:
https://docs.wagtail.io/en/v2.9/reference/pages/model_recipes.html#custom-tag-models


ababic:
Cheers @kevinhowbrook!

PR #426, opened 2020-06-02
Feature/p2 450 staff detail

kevinhowbrook:
[Ticket 450](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/450)
On dev site here https://rca-development.herokuapp.com/research-rca/dale-harrow/

PR #428, opened 2020-06-03
Feature/p2 478 staff roles

kevinhowbrook:
Depends on being merged first https://github.com/torchbox/rca-wagtail-2019/pull/426

[Ticket 478](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/478)

The questionable parts of this are the [role model itself](https://github.com/torchbox/rca-wagtail-2019/pull/428/files#diff-4b6ce2ef822805497c99bd9f0dc6d2bdR31) and the way I'm formatting the [data into groups](https://github.com/torchbox/rca-wagtail-2019/pull/428/files#diff-4b6ce2ef822805497c99bd9f0dc6d2bdR190)

The way I've done the formatting means that grouping will work with roles that have an actual programme_id relationship, and also ones that have a custom_programme (string)

roles-be

roles-fe


rca/people/models.py
```diff
@@ -1,65 +1,203 @@
...
+                    "description": page.introduction,
+                    "meta": meta,
+                }
+            )
+        return items
+
+    def get_roles_grouped(self):
+        items = self.role.all()
+        # TODO this query is way above my head
+        return items
```
kevinhowbrook, line 193: I've no idea how to do this query

rca/people/models.py
```diff
@@ -1,65 +1,219 @@
...
+        # Create a dictionary of values re-using keys so we can group by both
+        # the programmes and the custom programmes.
+        re_grouped = {}
+        for (key, value, link) in items:
+            if key in re_grouped:
+                re_grouped[key]["items"].append(value)
+            else:
+                re_grouped[key] = {"items": [value]}
+            re_grouped[key]["link"] = link
+        return re_grouped.items()
```
kevinhowbrook, line 209: I couldn't figure out how to write the query for this, so I had to do it by juggling the data around.

ababic: @kevinhowbrook Just wondering whether we really need this now? Seems like it would make more sense to obey the order imposed by Orderable, so the editor is in control?

rca/people/models.py
```diff
@@ -24,6 +28,37 @@ def __str__(self):
...
+        related_name="related_programme",
+        null=True,
+        blank=True,
+    )
+    custom_programme = models.CharField(
+        max_length=128,
+        help_text=_("Specify a custom programme page here if one does not exist"),
+        blank=True,
+    )
+    page = ParentalKey("StaffPage", related_name="role")
```
ababic, line 45: Since a staff member can have many roles, it might make more sense for the related name to be pluralised here.

rca/people/models.py
```diff
@@ -90,6 +125,9 @@ class StaffPage(BasePage):
             ],
             heading="Details",
         ),
+        MultiFieldPanel(
+            [InlinePanel("role", label=_("Staff role"))], heading=_("Staff roles")
+        ),
```
ababic, line 130: Is the MultiFieldPanel necessary here? I always thought InlinePanel looked fine as a top-level item.

rca/people/models.py
ababic: If we really do need this, I would personally do it at the database level, by annotating the queryset with a new string (the programme page title (if set), or the custom programme), then use that to apply ordering. Here's a really complex example from Wagtail:
https://github.com/wagtail/wagtail/blob/master/wagtail/core/sites.py#L14

This version would just need one `When` clause, plus the 'default'.

rca/people/models.py
kevinhowbrook: removed it now since we have a label specified

PR #430, opened 2020-06-03
[456] Feature legacy site news and events on research centre and landing pages

ababic:


.circleci/config.yml
```diff
@@ -60,7 +60,7 @@ jobs:
           name: Run Black
           command: |
             . venv/bin/activate
-            black --check ./rca
+            black --check ./rca --exclude migrations/
```
kevinhowbrook, line 63: yea good idea actually

rca/utils/models.py
```diff
@@ -373,6 +379,84 @@ def get_context(self, request, *args, **kwargs):
...
+        ),
+        through=LegacySiteTaggedPage,
+        blank=True,
+    )
+
+    class Meta:
+        abstract = True
+
+    @property
+    def legacy_news_cache_key(self):
```
kevinhowbrook, line 413: should this be `legacy_news_and_events_cache_key`? Seen as though it's news _and_ events that will be in the cache. Guess it's only called internally anyway so not a big deal

rca/landingpages/models.py
```diff
@@ -138,7 +144,7 @@ def __str__(self):
         return self.title
 
 
-class LandingPage(BasePage):
+class LandingPage(LegacyNewsAndEventsMixin, BasePage):
```
kevinhowbrook, line 147: I really like the way you've done this, it's going to be really useful down the line for other models

rca/utils/models.py
ababic: yes, it just seemed to be getting a bit too length/repetitive.

rca/landingpages/models.py
ababic: Cheers! I was tempted to stick this on BasePage, but that would have meant a hell of a lot of migrations :)

PR #442, opened 2020-06-10
Add some basic tests and allow for 404 responses

kevinhowbrook:


rca/people/models.py
```diff
@@ -226,8 +226,11 @@ def related_students_cache_key(self):
     def fetch_related_students(self):
         value = []
         if self.legacy_staff_id:
-            value = pull_related_students(self.legacy_staff_id)
-            cache.set(self.related_students_cache_key, value, None)
+            try:
+                value = pull_related_students(self.legacy_staff_id)
+                cache.set(self.related_students_cache_key, value, None)
+            except CantPullFromRcaApi:
+                pass
```
kevinhowbrook, line 233: - If a StaffPage is added with a bad legacy_staff_id, E.G `0` the api fetch responds with a 404, this was causing the page to break.

PR #444, opened 2020-06-10
[460] Anchor nav active state

chris-lawton:
https://projects.torchbox.com/projects/rca-website-rebuild/tickets/460

### DIFF
--- a/rca/static_src/javascript/components/anchor-nav.js
+++ b/rca/static_src/javascript/components/anchor-nav.js
@@ -1,0 +1,66 @@
+import 'intersection-observer';
+import scrollama from 'scrollama';
+
+class AnchorNav {
+    static selector() {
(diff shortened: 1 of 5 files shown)


rca/static_src/javascript/components/anchor-nav.js
```diff
@@ -0,0 +1,78 @@
...
+                }
+            })
+            .onStepExit((el) => {
+                if (el.direction === 'up') {
+                    this.handleUp(el);
+                }
+            });
+
+        // setup resize event
+        window.addEventListener('resize', scroller.resize);
```
SimonDEvans, line 37: We could use [`ResizeObserver`](https://web.dev/resize-observer/) here, but the performance gains may be minimal…

PR #450, opened 2020-06-12
WIP: 1069 - Site-wide overlay alert message

jafacakes2011:
I have currently done the following tasks:
 - Create a Site-wide alert setting in the Wagtail Admin
 - Conditionally output the alert message on all pages
    - This is now ready for FE to implement the alert box with the ability to close (and prevent appearing if previously closed)
 - Expose the Sitewide alert via an API endpoint `/api/v3/sitewide-alert/`

Tasks remaining:
 - [ ] Implement read and cache mechanism on the old site to read the alert from the rebuild site's API (won't be included as part of this PR)
 - [ ] FE integration


rca/utils/models.py
```diff
@@ -429,3 +430,21 @@ class Meta:
...
+class SitewideAlertSetting(BaseSetting):
+    class Meta:
+        verbose_name = "Sitewide alert"
+
+    show_alert = models.BooleanField(
+        default=False, help_text="Checking this will show the site-wide message"
+    )
+    message = RichTextField(
+        help_text="The message to be shown to all users across the site",
+        features=["h2", "h3", "bold", "italic", "ul", "ol", "link"],
```
kevinhowbrook, line 445: Could we remove UL OL please

jafacakes2011:
```suggestion
        features=["h2", "h3", "bold", "italic", "link"],
```

rca/project_styleguide/templates/patterns/molecules/sitewide-alert/sitewide-alert.html
```diff
@@ -0,0 +1,19 @@
+{% load wagtailcore_tags %}
+
+<div class="sitewide-alert" data-sitewide-alert data-sitewide-alert-close>
+    <button data-sitewide-alert-close class="sitewide-alert__close">
```
chris-lawton, line 4: can we add an `aria-label="Close alert"` to this?

nicklee:
This seems to have broken the js on staging, I suspect it's because someone disabled the message in wagtail, then the js couldn't find the data attribute in the DOM.

With that in mind I've applied a fix to not run the js if it doesn't exist.

PR #462, opened 2020-06-18
[1069] Site-wide overlay alert message (FE)

janehughes:
[Codebase ticket](https://projects.torchbox.com/projects/rca-django-cms-project/tickets/1069)

- [x] checked that this feature works across all browsers and devices we have agreed to support.
- [x] checked that this feature has data for the pattern library and does not break existing components in the pattern library
- [ ] checked that this code meets the accessibility level we have agreed to.
- [x] checked that this code meets the browser performance levels we have agreed to.
- [x] checked that this code conforms to the project coding standards.
- [ ] had code reviewed by another developer and resolved any issues raised.
- [x] tested this feature as an end-user.
- [ ] checked that the feature works on the server/s I am deploying it to.

### DIFF
--- a/rca/static_src/sass/components/_sitewide-alert.scss
+++ b/rca/static_src/sass/components/_sitewide-alert.scss
@@ -1,0 +1,55 @@
+.sitewide-alert {
+    @include z-index(modal);
+    display: none;
+    position: fixed;
+    left: 0;
+    top: 0;
+    width: 100vw;
+    height: 100vh;
+    overflow: auto;
+    background-color: $color--black-80;
+
+    // Applied if cookie just added
+    &--inactive {
+        display: none;
+    }
+
+    &--active {
+        display: block;
+    }
+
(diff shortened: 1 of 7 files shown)


rca/static_src/sass/components/_sitewide-alert.scss
```diff
@@ -0,0 +1,55 @@
...
+        position: relative;
+        top: 50%;
+        transform: translateY(-50%);
+        width: 100%;
+        max-width: 500px;
+        margin: 0 auto;
+        padding: ($gutter * 1.5);
+    }
+
+    &__heading {
```
nicklee, line 34: I can't see a sitewide-alert__heading class in the html, guess this might not be needed in the CSS?

janehughes: Good spot Nick! Thanks :)

PR #467, opened 2020-07-13
[508] Accessibility updates

chris-lawton:
https://projects.torchbox.com/projects/rca-website-rebuild/tickets/508

### DIFF
--- a/rca/project_styleguide/templates/patterns/atoms/modal-launcher/modal-launcher.html
+++ b/rca/project_styleguide/templates/patterns/atoms/modal-launcher/modal-launcher.html
@@ -5,7 +5,7 @@
     <a href="{{ item.link }}" class="modal-launcher">
 {% endif %}
     <div class="modal-launcher__container">
-        <div class="modal-launcher__label">
+        <div class="modal-launcher__label" id="{{ modal_aria_label }}">
             {{ item.action }}
         </div>
         <div class="modal-launcher__icon-container">
--- a/rca/project_styleguide/templates/patterns/molecules/navigation/primarynav.html
+++ b/rca/project_styleguide/templates/patterns/molecules/navigation/primarynav.html
@@ -35,6 +35,7 @@
                                 data-menu="{{ forloop.counter }}"
                                 data-menu-id={{ forloop.counter }}
                                 {% if link.secondary_links %}data-menu-parent{% endif %}
+                                role="menuitem"
                             >
                                 <span>{{ link.text }}</span>
                             </a>
@@ -49,6 +50,7 @@
                                     data-menu="{{ forloop.counter }}"
                                     data-menu-id={{ forloop.counter }}
                                     {% if link.secondary_links %}data-menu-parent{% endif %}
+                                    aria-label="{{ link.text }} submenu"
                                 >
                                     <svg class="nav__icon" width="15" height="15"><use xlink:href="#chevron"></use></svg>
                                 </a>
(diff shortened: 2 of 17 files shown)


rca/project_styleguide/templates/patterns/atoms/modal-launcher/modal-launcher.html
```diff
@@ -5,7 +5,7 @@
     <a href="{{ item.link }}" class="modal-launcher">
 {% endif %}
     <div class="modal-launcher__container">
-        <div class="modal-launcher__label">
+        <div class="modal-launcher__label" id="{{ modal_aria_label }}">
```
helenb, line 8: should this be an `aria-label` rather than an `id`?

rca/project_styleguide/templates/patterns/molecules/navigation/primarynav.html
```diff
@@ -81,6 +83,7 @@
                                     data-menu-id="{{ forloop.parentloop.counter}}-{{ forloop.counter }}"
                                     data-menu="{{ forloop.parentloop.counter }}-{{ forloop.counter }}"
                                     {% if item.tertiary_links %}data-menu-parent{% endif %}
+                                    role="menuitem"
```
helenb, line 86: same query as the other PR - do we definitely want to add these roles?

rca/project_styleguide/templates/patterns/molecules/profile-detail/profile-detail.html
```diff
@@ -21,9 +21,11 @@
 
     <div class="profile-detail__content">
         <div class="profile-detail__overview">
-            <div class="profile-detail__header">
-                <h3 class="profile-detail__heading heading heading--five">{{ item.first_name }} {{ item.surname }}</h3>
-            </div>
+            {% if item.first_name %}
```
helenb, line 24: should this test for both `item.first_name` and `item.surname`? or one or the other? I guess you don't want a surname on its own.

helenb:
Thanks @chris-lawton, just a couple of questions for you.

rca/project_styleguide/templates/patterns/molecules/navigation/primarynav.html
chris-lawton: I added them as the parent `<ul>`'s had a role of `menu` and Wave was reporting broken ARIA menu's without them. I hadn't realised these had all been removed from wagtail kit - what was the reason for that? I guess the other option is to remove all the roles from the menu entirely.

rca/project_styleguide/templates/patterns/atoms/modal-launcher/modal-launcher.html
chris-lawton: it should be `id`, just a confusing variable name

rca/project_styleguide/templates/patterns/molecules/profile-detail/profile-detail.html
chris-lawton: I thought that as well, but i think that it's a safe assumption that they wouldn't want to just show the surname

rca/project_styleguide/templates/patterns/molecules/navigation/primarynav.html
helenb: I think Thibaud said they were incorrect or unhelpful. Might be worth double checking with him if you have time, otherwise if it is ok now for the automated checks it should be fine to leave as is I think.

PR #472, opened 2020-08-04
Feature/p2 backend fixes

kevinhowbrook:
Commits pre-fixed with ticket numbers


rca/projects/models.py
```diff
@@ -326,11 +326,15 @@ def clean(self):
         if errors:
             raise ValidationError(errors)
 
-    def get_related_school(self):
-        """ returns the first related schools page"""
+    def get_related_school_or_centre(self):
+        # returns the first related schools page, if none, return the related research
+        # centre page
         realted_school = self.related_school_pages.first()
+        realted_research_page = self.related_research_pages.first()
```
nicklee, line 333: realted, is this a typo? Seems to be a consistent typo through if so :P

rca/project_styleguide/templates/patterns/pages/researchcentre/research_centre.html
```diff
@@ -85,7 +84,8 @@ <h2 class="section__heading section__heading--primary heading heading--four">
                         <div class="section__content grid">
                             {% include "patterns/organisms/carousel/carousel.html" with carousel=projects modifier='carousel carousel--no-margin' datatag='data-carousel' %}
                             <div class="section__action">
-                                <a class="link link--primary link--link body body--two" href="/research">
+                                <a class="link link--primary link--link body body--two" href="/research/projects/?school-or-centre={{ page.slug }}#results">
```
jafacakes2011, line 87: Hardcoded URLs worry me, Is this a URL that we can load using its name?
`/research/projects/`

kevinhowbrook: A valid point, there will only be one research section in the site, but I'm also a bit nervous about depending on these slugs, so I've swapped it for a URL field here in https://github.com/torchbox/rca-wagtail-2019/pull/472/commits/435b39400597d43153f950616b5cf6bf5a8b4c76

PR #480, opened 2020-10-01
1105 Allow publishers to choose guides and short courses here

kevinhowbrook:
Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/1105


jafacakes2011:
LGTM

PR #481, opened 2020-10-05
Add 'display' value field for the telephone number

kevinhowbrook:
Ticket https://projects.torchbox.com/projects/rca-django-cms-project/tickets/1091

- Opted for a new field here so the phonenumber field still gets
validated but the displayed value can be words or numbers in any format


rca/research/migrations/0015_update_help_text.py
```diff
@@ -0,0 +1,22 @@
...
+        ("research", "0014_tel_display_field"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="researchcentrepage",
+            name="centre_tel_display_text",
+            field=models.CharField(
+                blank=True,
+                help_text="Specify specific text or numbers to display for the linked tel number, e.g. +44 (0)20 7590 1234 or +44 (0)7749 183783",
```
kevinhowbrook, line 18: This has gone to staging, otherwise, I would regenerate this as one migrations

tombola:
**:thought:** In general, looking at the ticket I am not entirely clear that I understand the approach, as it looks like the displayed number might diverge from that provided as the actual phone number, and also if the styleguide dictates a specific pattern for phone numbers, which are also validated on submission, I would have anticipated the display format could have been accomplished by template filter. I think I might have missed something in this regard.

If this has been considered and discounted already, the actual implementation looks fine to me.

PR #482, opened 2020-10-15
Add the option to disable logging here as it's making sentry too noisy.

kevinhowbrook:


jafacakes2011:
Changes look good to me 😄

PR #485, opened 2020-10-27
Adding booking bar to programme page

nicklee:
Adding the booking bar component to the programme page


nicklee:
This is still blocked pending design review: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/1077

PR #507, opened 2020-12-15
Stop draft pages appearing in react listings

kevinhowbrook:


richbrennan:
Looks good. One suggestion, is it possible to apply the `live()` filter to the `queryset` that gets passed in, like in `BaseFilterBackend`?

kevinhowbrook:
> Looks good. One suggestion, is it possible to apply the `live()` filter to the `queryset` that gets passed in, like in `BaseFilterBackend`?

Thanks rich, I can't quite see a way to override that before, as filter_queryset must be overridden https://github.com/encode/django-rest-framework/blob/d635bc9c71b6cf84d137a68610ae2e628f8b62b3/rest_framework/filters.py#L28

PR #514, opened 2021-01-05
608 school news and events

kevinhowbrook:
*contains* https://github.com/torchbox/rca-wagtail-2019/pull/511

Ticket [608](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/608)

This pull request:
 - Enables News and Event listing by tag to the legacy site. 
 - Defines a carousel to be used for student stories

PR #512, opened 2021-01-04
651 enable wagtail forms

kevinhowbrook:
Ticket [651](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/651)

Small change to enable Wagtail forms

PR #515, opened 2021-01-06
610 contact

kevinhowbrook:
**Contains #514 and #512**
Ticket [610](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/610)

In summary, this PR implements a new mixin in utils.models as ContactFieldsMixin. We are introducing the ability to link to a 'contact form' offered by a page chooser. In an effort to reduce the repeated code, the mixin will help us keep the validation in one place.

Migrations process:

1. Add new fields via mixin as `contact_model_[name]`
2. Migrate data from old contact fields to new fields (see utils/migrations/0019)
3. Remove the old contact fields

Caveats
- Not all models use _all_ the fields offered by the new mixin, mainly this is the 'contact title', in this instance the panel isn't exposed.

@richbrennan due to the branch dependencies here, it might be easier to review this together or at least wait for a rebase


richbrennan:
Some conflicts to resolve

kevinhowbrook:
> Some conflicts to resolve

Thanks @richbrennan I've reset this against p3-schools integration branch so now only the relevant diff is shown. It does look like a lot still as of the migrations needed to cleanup the contact code into a mixin [mention at the top of this](https://github.com/torchbox/rca-wagtail-2019/pull/515#issue-550351362)

richbrennan:
@kevinhowbrook 

So is this waiting for any more merges/rebases?

kevinhowbrook:
> @kevinhowbrook
> 
> So is this waiting for any more merges/rebases?

Nope, this one is all new additions

rca/utils/models.py
```diff
@@ -664,3 +666,66 @@ class Sector(SluggedTaxonomy):
     class Meta:
         verbose_name = "Innovation RCA sector"
         verbose_name_plural = "Innovation RCA sectors"
+
+
+class ContactFieldsMixin(models.Model):
```
richbrennan, line 671: All the fields on this model can be blank, should at least one (maybe title) be required?

richbrennan (commented):
Just one comment, thanks Kevin

rca/utils/models.py
kevinhowbrook: It's a good point, we handle a default title value field in the template https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/project_styleguide/templates/patterns/organisms/contact/contact.html#L9

The snag here is that some pages _do_ have a contact title, and some don't, and contact details at the page level aren't required so I think this is the best I can get out of something re-usable like this.

PR #521, opened 2021-01-11
Stop logging everything

kevinhowbrook:
Ticket [656](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/656)

The get request for the social image should go through the custom `fetch` method since it passes through the catching pattern. At the moment it's not, which is resulting in [ReadTimeout](https://sentry.io/organizations/torchbox/issues/?environment=production&project=1542908&query=is%3Aunresolved&statsPeriod=24h) exceptions being raised.

This will mean `CantPullFromRcaApi` will be raised, ideally, I'd like to completely silence this logging with the `settings.API_FETCH_LOGGING` env var I set up (the legacy site isn't very responsive to the requests) but they are still occasionally being logged in sentry

### DIFF
--- a/rca/api_content/content.py
+++ b/rca/api_content/content.py
@@ -100,8 +100,11 @@ def parse_items_to_list(data, type):
 
         if "social_image" in data and data["social_image"]:
             social_image = data["social_image"]["meta"]["detail_url"]
-            social_image = requests.get(url=social_image, timeout=10)
-            social_image = social_image.json()
+            social_image = fetch_data(url=social_image)
+            try:
+                social_image = social_image.json()
+            except AttributeError:
+                continue
             if "url" in social_image["rca2019_feed_image"]:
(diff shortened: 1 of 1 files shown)


rca/api_content/content.py
```diff
@@ -100,8 +100,11 @@ def parse_items_to_list(data, type):
 
         if "social_image" in data and data["social_image"]:
             social_image = data["social_image"]["meta"]["detail_url"]
-            social_image = requests.get(url=social_image, timeout=10)
-            social_image = social_image.json()
+            social_image = fetch_data(url=social_image)
+            try:
+                social_image = social_image.json()
```
richbrennan, line 105: Would it be worth calling `.json()` in the  `fetch_data()` function before its returned? This would keep calling code ignorant of a data structure it probably doesn't need to know about?

richbrennan (commented):
One comment, thanks Kevin!

PR #528, opened 2021-01-13
Feature/620 student research

kevinhowbrook:
**contains https://github.com/torchbox/rca-wagtail-2019/pull/525**
Ticket [620](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/620#update-65778107)


rca/people/formatters.py
```diff
@@ -0,0 +1,23 @@
...
+        items.append(
+            {
+                "title": page.title,
+                "link": page.url,
+                "image": page.hero_image,
+                "description": page.introduction,
+                "meta": meta,
+            }
+        )
+    return items
```
kevinhowbrook, line 23: Shifted this to prevent duplicating it for both students and staff pages

rca/people/utils.py
```diff
@@ -0,0 +1,46 @@
+from django.apps import apps
```
kevinhowbrook, line 1: The functions here are based off the existing work in https://github.com/torchbox/rca-wagtail-2019/blob/84ad028150dbc961fcf8e240e982ab30998861b2/rca/people/models.py#L216-L235

Rather than duplicating them entirely in the model, I've moved them here; with an aim to making them one function/class

rca/people/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    # First return any editorially-highlighted project pages
+    for p in page.related_project_pages.all():
+        related_project_page_ids.append(p.page.id)
+        yield p.page.specific
+
+    # Then return any other project pages which the staff member leads or is a team member of,
+    # filtering out any of the highlights already output
+    yield from ProjectPage.objects.filter(
+        Q(project_lead__page_id=page.pk) | Q(related_staff__page_id=page.pk)
+    ).exclude(pk__in=related_project_page_ids).order_by(
```
kevinhowbrook, line 22: This is really the only specific part of this function, meaning I had to duplicate it below.

Perhaps we can pass in the Q() values as an argument?

rca/people/formatters.py
```diff
@@ -0,0 +1,23 @@
+def format_research_highlights(pages):
+    """Function for formatting related projects to the correct
+    structure for the gallery template
+    Returns:
+        List
+    """
+    items = []
+    for page in pages:
```
richbrennan, line 8: would `select_related('related_school_pages')` or `prefetch_related('related_school_pages')` help reduce queries?

richbrennan (commented):
Some of this looks familiar, does other work need to be merged in?

rca/people/formatters.py
kevinhowbrook: I don't think I can do that, the reason being is that the research pages [here](https://github.com/torchbox/rca-wagtail-2019/blob/cf8b80e814a69d6c1d2068b5300f7494912661b5/rca/people/models.py#L338-L339) come from a generator (yield) in https://github.com/torchbox/rca-wagtail-2019/blob/cf8b80e814a69d6c1d2068b5300f7494912661b5/rca/people/utils.py#L1-L46

kevinhowbrook:
> Some of this looks familiar, does other work need to be merged in?

Yea some of this is similar to what's in the other p3-schools branch, formatters, but for different models.

rca/people/utils.py
```diff
@@ -0,0 +1,46 @@
+from django.apps import apps
+from django.db.models import Q
+
+
+def get_staff_research_projects(page):
+    """Yields a list combining project pages editorially-selected on the staff page,
```
richbrennan, line 6: It doesn't yield a list, it yields pages

rca/people/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    and those on which the the staff member is listed as lead or a team member
```
richbrennan, line 7: Repeated `the`

rca/people/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    # filtering out any of the highlights already output
+    yield from ProjectPage.objects.filter(
+        Q(project_lead__page_id=page.pk) | Q(related_staff__page_id=page.pk)
+    ).exclude(pk__in=related_project_page_ids).order_by(
+        "-first_published_at"
+    ).distinct()
+
+
+def get_student_research_projects(page):
+    """Yields a list combining project pages editorially-selected on the staff page,
```
richbrennan, line 28: see above

rca/people/utils.py
```diff
@@ -0,0 +1,46 @@
...
+    and those on which the the staff member is listed as lead or a team member
```
richbrennan, line 29: repeated `the`

richbrennan (commented):
A few minor points, thanks Kevin!

PR #537, opened 2021-01-20
Allow new school pages to be viewable by is_staff for publishing

kevinhowbrook:
So publishing can begin once we deploy the work... and so that we can preview the new pages, this PR adds a 404 response set on the schools serve method depending on the user.

All is_staff users should be able to see the new schools pages for publishing purposes, but the public should receive a 404 and be rerouted to the old site accordingly until the pages are ready. Eventually, when ready,  we'll remove the custom serve method altogether.

### DIFF
--- a/rca/schools/models.py
+++ b/rca/schools/models.py
@@ -4,6 +4,8 @@ from collections import defaultdict
 from django.apps import apps
 from django.core.exceptions import ValidationError
 from django.db import models
+from django.http import Http404
+from django.shortcuts import render
 from django.utils.translation import ugettext_lazy as _
 from modelcluster.fields import ParentalKey
 from wagtail.admin.edit_handlers import (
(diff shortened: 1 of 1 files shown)


rca/schools/models.py
```diff
@@ -176,6 +178,13 @@ def __str__(self):
 
 
 class SchoolPage(ContactFieldsMixin, LegacyNewsAndEventsMixin, BasePage):
+    # Allow staff to see the new school pages for publishing, anyone else
+    # should receive a 404 and get redirected to the legacy site.
+    def serve(self, request):
+        if not request.user.is_staff:
+            raise Http404
+        return render(request, self.template, self.get_context(request))
```
richbrennan, line 186: This is different to the Wagtail `Page` implementation, which takes into account previews, do you really want to override it? https://github.com/wagtail/wagtail/blob/master/wagtail/core/models.py#L1600

You could defer to default Wagtail behaviour with `return super().serve(request)`

richbrennan (commented):
One comment, thanks Kevin

PR #543, opened 2021-01-22
Stop allowing blank page references

kevinhowbrook:


rca/guides/migrations/0010_stop_allowing_blank_pages.py
```diff
@@ -0,0 +1,24 @@
...
+        ("guides", "0009_allow_blank_values"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="guidepagerelatedpages",
+            name="page",
+            field=models.ForeignKey(
+                null=True,
+                on_delete=django.db.models.deletion.CASCADE,
```
kevinhowbrook, line 19: I'm in two minds weather to set the to PROTECT...

richbrennan: I think CASCADE is fine because this is just a related page placement. If the page needs to be deleted, it'd be annoying to have to go into every page where its used to unset. I think it'd 500 error with PROTECT anyway so the user wouldn't even be informed which ones.

richbrennan (commented):
Have replied to your question, otherwise good

PR #545, opened 2021-02-08
Fixing OG images

nicklee:


nicklee:
Will be on staging in 10: https://rca-staging.herokuapp.com/study/courses/architecture-ma/

chris-lawton (approved):

PR #565, opened 2021-02-18
#624 BE Enquire to study Form - QS integration

BrianXu20:
Ticket:
https://projects.torchbox.com/projects/rca-website-rebuild/tickets/624


kevinhowbrook (commented):
Thanks @BrianXu20 this looks good to me.

I'm conscious we don't have all the field data from the form in the post data here, this is large because we don't know what those are. I'll get a response together for this but we should wait until we have all the data before merging this in.

kevinhowbrook (changes requested):

rca/enquire_to_study/views.py
```diff
@@ -28,7 +30,23 @@ def post_mailchimp(self):
...
     def post_qs(self, form_data):
-        pass
+        # send data to QS student enquiry endpoint
+
+        data = {
+            "FirstName": form_data["first_name"],
+            "LastName": form_data["last_name"],
+            "EmailAddress": form_data["email"],
+            "MobileNumber": f"{form_data['phone_number'].country_code}||{form_data['phone_number'].national_number}",
+        }
```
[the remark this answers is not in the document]
BrianXu20: On the form, LevelOfStudy and Course are multiple choices, I am not sure if this API accepts multiple entries.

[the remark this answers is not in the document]
BrianXu20: I think I used the wrong field for LevelOfStudy, it should be the level of study of the programmes.
But the programmes still remain multiple choices,
The possible solutions are :
1. Comma separated 
2. Multiple records (preferable )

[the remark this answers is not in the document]
kevinhowbrook: I think Comma separated values will be ok for this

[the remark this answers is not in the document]
BrianXu20: The code is revised for this decision.
Thanks Kevin

rca/enquire_to_study/views.py
```diff
@@ -28,7 +30,71 @@ def post_mailchimp(self):
...
+        # Encode level of study and course in QS API
+        # Documentation http://qs-enrolment-solutions.screenstepslive.com/s/Help/m/enquiryapi/l/889815-application-endpoints
+
+        selected_level_of_study_list = []
+        selected_course_list = []
+        for programme in form_data["programmes"]:
+            course_code = next(
+                (course for course in qs_courses if course["name"] == programme.title),
+                None,
+            )
```
kevinhowbrook, line 60: Nice, I like how you've done this :clap:

BrianXu20: Haha, thanks

rca/enquire_to_study/views.py
```diff
@@ -18,22 +52,112 @@ def dispatch(self, *args, **kwargs):
...
+        for programme in form_data["programmes"]:
+            course_code = next(
+                (
+                    course
+                    for course in qs_courses
+                    if course["codeExternal"] == str(programme.qs_code)
+                ),
+                None,
+            )
+            if not course_code:
```
richbrennan, line 112: If `None` is removed from the call to `next()` it'll raise a `StopIteration` exception if the programme isn't found, so the `if` check could be removed and `StopIteration` caught instead.

rca/enquire_to_study/views.py
```diff
@@ -18,22 +52,112 @@ def dispatch(self, *args, **kwargs):
...
+                raise Exception(f"{programme.title} not found in QS course list")
```
richbrennan, line 113: Would it be worth raising a more specific exception, maybe `ValueError`? https://docs.python.org/3/library/exceptions.html

rca/enquire_to_study/views.py
```diff
@@ -18,22 +52,112 @@ def dispatch(self, *args, **kwargs):
...
+        data["Course"] = ",".join(selected_course_list)
+        response = requests.post(
+            f"{settings.QS_API_ENDPOINT}/studentEnquiries",
+            json=data,
+            auth=(settings.QS_API_USERNAME, settings.QS_API_PASSWORD),
+        )
+        # If the response was successful, no Exception will be raised
+        response.raise_for_status()
 
     def create_form_submission(self, form):
```
richbrennan, line 155: Where is `create_form_submission()` called? I wonder if the indirection to call `form.save()` is necessary

rca/enquire_to_study/views.py
```diff
@@ -18,22 +52,112 @@ def dispatch(self, *args, **kwargs):
         return super().dispatch(*args, **kwargs)
 
     def post_mailchimp(self):
```
richbrennan, line 54: This method is called below but doesn't do anything, might it be worth raising `NotImplemented`, or add a comment about what needs to added?

rca/programmes/models.py
```diff
@@ -92,6 +92,12 @@ class ProgrammePageSubjectPlacement(models.Model):
 class ProgrammeType(WagtailOrdable):
     display_name = models.CharField(max_length=128)
     description = models.CharField(max_length=500, blank=True)
+    qs_code = models.CharField(
+        max_length=500,
+        help_text="This code needs to match the name of the LevelOfStudy code value in QS",
+        blank=True,
+        null=True,
```
richbrennan, line 99: Null not needed for CharField

rca/programmes/models.py
```diff
@@ -417,6 +423,11 @@ class ProgrammePage(ContactFieldsMixin, BasePage):
         ],
         blank=True,
     )
+    qs_code = models.PositiveIntegerField(
+        help_text="This code needs to match the name of the codeExternal value in QS, E.G 105",
+        blank=True,
+        null=True,
```
richbrennan, line 429: Null not needed for CharField

richbrennan (commented):
Thanks Kevin, a few comments

rca/enquire_to_study/views.py
kevinhowbrook: That's called here https://github.com/torchbox/rca-wagtail-2019/blob/feature/624-qs-integration/rca/enquire_to_study/views.py#L195 and the from.save creates the form submission object. 

Is the suggestion here that I just call `form.save()` [here](https://github.com/torchbox/rca-wagtail-2019/blob/feature/624-qs-integration/rca/enquire_to_study/views.py#L195) instead?

rca/enquire_to_study/views.py
kevinhowbrook: added a comment

rca/programmes/models.py
kevinhowbrook: This one is an integer field, and I need it to be null to control form options like this https://github.com/torchbox/rca-wagtail-2019/blob/feature/624-qs-integration/rca/enquire_to_study/forms.py#L36

```
programmes = forms.ModelMultipleChoiceField(
        queryset=ProgrammePage.objects.filter(qs_code__isnull=False, live=True),
        widget=forms.CheckboxSelectMultiple,
    )
```

PR #567, opened 2021-02-19
1092 Ensure these pages are live before deciding to list them

kevinhowbrook:
contains security updates and treebeard pin so the pipeline passes

### DIFF
--- a/rca/people/models.py
+++ b/rca/people/models.py
@@ -246,10 +246,9 @@ class StaffPage(BasePage):
         related_project_page_ids = []
 
         # First return any editorially-highlighted project pages
-        for p in self.related_project_pages.all():
-            if p.page.live:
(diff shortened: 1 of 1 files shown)


rca/people/models.py
```diff
@@ -247,16 +247,17 @@ def get_research_projects(self):
 
         # First return any editorially-highlighted project pages
         for p in self.related_project_pages.all():
-            related_project_page_ids.append(p.page.id)
-            yield p.page.specific
+            if p.page.live:
```
richbrennan, line 250: You could move this filter to the original query:

```
for p in self.related_project_pages.filter(page__live=True):
    ...
```

richbrennan (commented):
One small recommendation, thanks Kevin!

PR #569, opened 2021-02-19
No longer wait for CSS on modal launch

nicklee:
Was causing issues when opening the modal for a second time


nicklee:
Fix for bug here https://projects.torchbox.com/projects/rca-website-rebuild/tickets/567

chris-lawton (approved):

PR #574, opened 2021-02-23
1131-embed-video-helptext

tombola:
https://projects.torchbox.com/projects/rca-django-cms-project/tickets/1131

I ended up squashing some migrations because of amendments after staging. 

Unsquashed migrations already applied on staging, but this change has passed acceptance and they would be untidy on prod so removed.


rca/utils/blocks.py
```diff
@@ -288,7 +288,10 @@ class GuideBlock(blocks.StreamBlock):
     paragraph = blocks.RichTextBlock()
     image = ImageBlock()
     quote = QuoteBlock()
-    embed = EmbedBlock()
+    embed = EmbedBlock(
+        label="Embed media",
+        help_text="Add a URL from these providers: YouTube, Vimeo, SoundCloud, Twitter.",
```
kevinhowbrook, line 293: **nitpick**

```suggestion
        help_text="Add a URL from one of these providers: YouTube, Vimeo, SoundCloud, Twitter.",
```

kevinhowbrook (approved):

PR #575, opened 2021-02-24
1132 - document helptext

tombola:
https://projects.torchbox.com/projects/rca-django-cms-project/tickets/1132

Adds a (recommended) maximum filesize for document uploads.

Took more investigation time than I probably should have to track this down to AbstractDocument model itself (and so to CustomDocument):

1. https://github.com/wagtail/wagtail/blob/423aefc72c1059121ebf4b8d517a47a0a0b393d6/wagtail/documents/templates/wagtaildocs/multiple/add.html#L21
2. https://github.com/wagtail/wagtail/blob/133e2b7a3742e059ec9b2748933193b5442e081f/wagtail/admin/views/generic/multiple_upload.py#L174
3. https://github.com/wagtail/wagtail/blob/b13641e9367b62eaedef85c8900bbb88efdf9354/wagtail/documents/models.py#L28


tombola:
Spotted that this poor request is still open, so have replaced it and re-created migrations.

PR #577, opened 2021-02-25
1172 - Amendments to hardcoded short course wording

tombola:
https://projects.torchbox.com/projects/rca-django-cms-project/tickets/1172


tombola:
This would be a waste of @kevinhowbrook 's time to review, as effectively reverting a previous commit, so merging now.

PR #584, opened 2021-03-08
#685 Re-implement specific fields for student information

kevinhowbrook:
Ticket [685](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/685)

This PR reworks some fields on the StudentPage model so we can render them on a form for users to update their own page. The rationale is that streamfield on a Django view/form won't behave well so it's better to be more specific where we can


rca/people/models.py
```diff
@@ -504,6 +504,21 @@ class RelatedStudentPage(Orderable):
     panels = [PageChooserPanel("page")]
 
 
+class StudentPageGallerySlide(Orderable):
+    source_page = ParentalKey("StudentPage", related_name="gallery_slides")
+    title = models.CharField(max_length=120)
+    image = models.ForeignKey(
```
richbrennan, line 510: Should image be `blank=True`?

rca/project_styleguide/templates/patterns/molecules/accordion/accordion.html
```diff
@@ -40,7 +40,7 @@ <h5 class="accordion__heading heading heading--five">{{ accordion.heading }}</h5
                     {{ accordion.introduction }}
                 </div>
             {% endif %}
-            {{ accordion.body|richtext }}
+            {{ accordion.body|richtext|linebreaks }}
```
richbrennan, line 43: Is `linebreaks` needed for existing data...?

rca/people/models.py
```diff
@@ -660,13 +700,83 @@ def name(self):
...
+                    }
+                }
+            )
+        return data
+
+    @property
+    def student_gallery(self):
+        # Format related model to a nice dict
+        data = []
+        if not self.gallery_slides.exists():
```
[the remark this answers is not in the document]
kevinhowbrook: Good point, I've removed it, the `for item in self.gallery_slides.all():` will handle it. I have an unhealthy fixation on trying to return early :rofl:

rca/people/models.py
```diff
@@ -660,13 +700,83 @@ def name(self):
...
+                        "author": item.author,
+                        "image": item.image,
+                    }
+                }
+            )
+        return data
+
+    @property
+    def student_related_links(self):
+        if not self.relatedlinks.exists():
```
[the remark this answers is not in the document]
kevinhowbrook: removed

rca/project_styleguide/templates/patterns/molecules/accordion/accordion.html
kevinhowbrook: Linebreaks is actually added to handle these new fields, as they aren't rich text. But the accordion is used allover the place with rich text fields.

PR #588, opened 2021-03-12
622 Fixes to initial values

kevinhowbrook:
This will need rebasing once #565 is in (it will fix tests)

### DIFF
--- a/rca/enquire_to_study/forms.py
+++ b/rca/enquire_to_study/forms.py
@@ -115,15 +115,14 @@ class EnquireToStudyForm(forms.Form):
 
     def save(self):
         data = self.cleaned_data.copy()
-        programme_types = data.pop("programme_types")
+        programme_type = data.pop("programme_types")
         programmes = data.pop("programmes")
         data.pop("captcha")
         enquiry_submission = EnquiryFormSubmission.objects.create(**data)
 
-        for programme_type in programme_types:
-            EnquiryFormSubmissionProgrammeTypesOrderable.objects.create(
-                enquiry_submission=enquiry_submission, programme_type=programme_type
-            )
+        EnquiryFormSubmissionProgrammeTypesOrderable.objects.create(
+            enquiry_submission=enquiry_submission, programme_type=programme_type
+        )
 
         for programme in programmes:
             EnquiryFormSubmissionProgrammesOrderable.objects.create(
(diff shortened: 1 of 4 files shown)


rca/enquire_to_study/forms.py
```diff
@@ -57,11 +60,10 @@ class EnquireToStudyForm(forms.Form):
...
         super(EnquireToStudyForm, self).__init__(*args, **kwargs)
         # Placeholder
-        self.fields["first_name"].widget.attrs["placeholder"] = "First name *"
-        self.fields["last_name"].widget.attrs["placeholder"] = "Last name *"
-        self.fields["email"].widget.attrs["placeholder"] = "Email *"
-        self.fields["phone_number"].widget.attrs["placeholder"] = "Phone number *"
-        self.fields["city"].widget.attrs["placeholder"] = "City or town of residence *"
+
+        # Set initial values
+        self.fields["country_of_residence"].initial = ("GB", "United Kingdon")
```
BrianXu20, line 65: This can be done sitewide and declarative:

Show certain countries first
Provide a list of country codes as the COUNTRIES_FIRST setting and they will be shown first in the countries list (in the order specified) before all the alphanumerically sorted countries.

If you want to sort these initial countries too, set the COUNTRIES_FIRST_SORT setting to True.

By default, these initial countries are not repeated again in the alphanumerically sorted list. If you would like them to be repeated, set the COUNTRIES_FIRST_REPEAT setting to True.

Finally, you can optionally separate these ‘first’ countries with an empty choice by providing the choice label as the COUNTRIES_FIRST_BREAK setting.

https://pypi.org/project/django-countries/#customize-the-country-list

rca/enquire_to_study/forms.py
```diff
@@ -27,9 +28,10 @@ class EnquireToStudyForm(forms.Form):
     country_of_citizenship = CountryField().formfield()
 
     # Study details
-    programme_types = forms.ModelMultipleChoiceField(
+    programme_types = forms.ModelChoiceField(
```
BrianXu20, line 31: Feels like programme_type makes more sense rather than programme_types

BrianXu20 (approved):
Left some small suggestions,
no major issue found

rca/enquire_to_study/forms.py
kevinhowbrook: Thanks @BrianXu20, This isn't regarding the order of the values, it's so the field loads with the initial value already selected:

image

PR #603, opened 2021-03-25
675 student key details additions

kevinhowbrook:
Ticket [675](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/675)


rca/people/models.py
```diff
@@ -686,6 +740,33 @@ def name(self):
         parts = (self.student_title, self.first_name, self.last_name)
         return " ".join(p for p in parts if p)
 
+    @property
+    def supervisors(self):
+        if not self.related_supervisor:
```
richbrennan, line 745: This check is probably redundant, as if there isn't a supervisor it'll return an empty list

rca/people/models.py
```diff
@@ -686,6 +740,33 @@ def name(self):
...
+                    }
+                )
+            else:
+                supervisors.append(
+                    {
+                        "title": item.title
+                        + " "
+                        + item.first_name
+                        + " "
+                        + item.surname,
```
richbrennan, line 763: An `f""` string might be a little nicer than all the `+`s

richbrennan (approved):

PR #608, opened 2021-04-06
#681 enquiry form tracking

kevinhowbrook:
Part 2 of https://projects.torchbox.com/projects/rca-website-rebuild/tickets/681

For the custom enquiry form, we need to ensure the form submission is sent to analytics.

### DIFF
--- a/rca/enquire_to_study/views.py
+++ b/rca/enquire_to_study/views.py
@@ -326,10 +326,12 @@ class EnquireToStudyFormThanksView(TemplateView):
 
         # If there is a session variable containing form post data send this,
         # through to the thanks template.
+        from django.utils.html import mark_safe
+
         if "enquiry_form_session_data" in self.request.session:
-            context["enquiry_form_session_data"] = self.request.session[
-                "enquiry_form_session_data"
-            ]
+            context["enquiry_form_session_data"] = mark_safe(
+                self.request.session["enquiry_form_session_data"]
+            )
             # In case the page is refreshed, make sure we clear the session form data.
             del self.request.session["enquiry_form_session_data"]
         return context
--- a/rca/enquire_to_study/views.py
+++ b/rca/enquire_to_study/views.py
@@ -260,9 +260,6 @@ class EnquireToStudyFormView(FormView):
         # If the response was successful, no Exception will be raised
         response.raise_for_status()
 
-    def create_form_submission(self, form):
-        form.save()
-
     def send_user_email_notification(self, form):
         enquiry_form_settings = EnquireToStudySettings.for_request(self.request)
         if not enquiry_form_settings.email_submission_notifations:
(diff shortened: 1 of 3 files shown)


rca/enquire_to_study/views.py
```diff
@@ -189,14 +189,32 @@ def send_user_email_notification(self, form):
             fail_silently=False,
         )
 
+    def set_session_data(self, form_data, enquiry_submission):
+        # Sets form data into session variable for analytics.
+        self.request.session["enquiry_form_session_data"] = {
+            "event": "form-submission",
+            "formName": "Enquire to study form",
+            "formId": enquiry_submission.id,
```
kevinhowbrook, line 197: In the ticket, this is requested as `'formId':'###',` - I'm waiting on a response from Chris danks, but the submission object ID that's created make most sense to me

rca/enquire_to_study/views.py
```diff
@@ -189,14 +189,32 @@ def send_user_email_notification(self, form):
...
+
     def form_valid(self, form):
         country_of_residence = form.cleaned_data["country_of_residence"]
         if country_of_residence in ["GB", "IE"]:
             self.post_mailchimp()
         else:
             self.post_qs(form.cleaned_data)
 
-        self.create_form_submission(form)
+        enquiry_submission = self.create_form_submission(form)
```
kevinhowbrook, line 213: I'm not a huge fan of the way this reads, or if it's bad practice. `create_form_submission` just calls `form.save()` so I'm just grabbing what I need by returning the [form.save()](https://github.com/torchbox/rca-wagtail-2019/pull/608/files#diff-52c257a033b79d236a9b23e7a64206ff76ae355021264e863d1083791b1eb889R165)

richbrennan: Yeah I'm not sure what benefit `create_form_submission` is, seems like a method just to call another method on the same class?

rca/enquire_to_study/views.py
```diff
@@ -189,14 +189,32 @@ def send_user_email_notification(self, form):
...
+        self.request.session["enquiry_form_session_data"] = {
+            "event": "form-submission",
+            "formName": "Enquire to study form",
+            "formId": enquiry_submission.id,
+            "countryResidence": form_data["country_of_residence"],
+            "countryCitizen": form_data["country_of_citizenship"],
+            "programme": [i.title for i in form_data["programmes"]],
+            "starting": form_data["start_date"].label,
+            "enquiryType": form_data["enquiry_reason"].reason,
+            "newsletter": True,
```
richbrennan, line 203: Are they being automatically signed up to a newsletter, or does this maybe need a dynamic value?

richbrennan (approved):
A couple of minor comments, thanks Kevin

rca/enquire_to_study/views.py
kevinhowbrook: :medal_military: thanks

rca/enquire_to_study/views.py
kevinhowbrook: I think in this instance it's turned out handy so I can retrieve the submission object created in form.save()

richbrennan: Would `enquiry_submission = form.save()` not work?

kevinhowbrook: :facepalm: good idea

rca/project_styleguide/templates/patterns/pages/enquire_to_study/enquire_form_thanks.html
```diff
@@ -33,5 +33,15 @@ <h1 class="title-area__heading heading heading--display-two">
...
     </div>
-
 {% endblock %}
+
+{% block extra_js %}
+    {% if GOOGLE_TAG_MANAGER_ID and enquiry_form_session_data %}
+        <script>
+            window.dataLayer = window.dataLayer || [];
+            window.dataLayer.push(
+                {{ enquiry_form_session_data|safe }}
```
richbrennan, line 43: Applying the `safe` filter to form data could pose a security risk. If its causing JS errors it might be worth changing the format of the data.

richbrennan (changes requested):

rca/project_styleguide/templates/patterns/pages/enquire_to_study/enquire_form_thanks.html
kevinhowbrook: I've marked this as safe before the template https://github.com/torchbox/rca-wagtail-2019/pull/608/commits/d5668e341a0c8d00558148209992ab2b4a516d66

PR #610, opened 2021-04-08
#706 Implements new filter to be used to force all rich text links to be _target=blank

kevinhowbrook:
[Ticket 706](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/706)

Needs to go after https://github.com/torchbox/rca-wagtail-2019/pull/604


rca/utils/templatetags/util_tags.py
```diff
@@ -179,3 +182,37 @@ def slice_pagination(page_range, current_page):
...
+
+    The main difference being we want to force all links to open in a new window, this is useful for front end
+    form elements where the user would risk losing any form data. Typical scenario is getting the the end of
+    a form, clicking a link to view "terms and conditions", then upon returning to the form page the form inputs
+    could be reset.
+
+    We still set html via `expand_db_html`, to take care of any processing that may be added already, or in the future.
+    """
+    if isinstance(value, RichText):
+        # passing a RichText value through the |richtext_force_external_links filter should have no effect
```
richbrennan, line 201: Why no effect, what if a RichText value is passed that contains links which need to have target set to blank?

richbrennan (changes requested):
One question, otherwise looks good, thanks Kevin.

rca/utils/templatetags/util_tags.py
kevinhowbrook: This is very much copied/based on the behaviour of the core filter https://github.com/wagtail/wagtail/blob/571b9e1918861bcbe64a9005a4dc65f3a1fe7a15/wagtail/core/templatetags/wagtailcore_tags.py#L98-L100

To my understanding, I only want to handle string values passed to the filter, not actual RichText objects.

richbrennan: OK, I'm not familiar with the `RichText` class, just wanted to flag that it wasn't being handled in case it needed to be

PR #613, opened 2021-04-12
Allowing modal users who use the back button to scroll the page

nicklee:


rca/static_src/javascript/components/modal.js
```diff
@@ -17,6 +17,11 @@ document.addEventListener('DOMContentLoaded', () => {
         const modalId = document.querySelector('[data-micromodal-trigger]')
             .dataset.micromodalTrigger;
         MicroModal.close(modalId);
+        // Make sure body is scrollable when modal is closed
+        document
+            .querySelector('body')
+            .style.removeProperty('overflow')
+            .removeProperty('height');
```
chris-lawton, line 24: I don't think you can chain the `.removeProperty` method - they both need to be called on the `.style` object in order to remove both. Something like this should work:

```
const bodyStyle = document.querySelector('body').style;
bodyStyle.removeProperty('overflow');
bodyStyle.removeProperty('height');
```

nicklee: Thanks dude!

PR #618, opened 2021-04-13
727 student profiles admin journey

kevinhowbrook:
Ticket [727](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/727)

image

- [x] Add migration to create new student role and permission
- [x] Add student_page__student_account relationship
- [x] Validation: Can only add 1 student to 1 student page (no student user can have multiple pages)
- [x] Signal to create a student page if one doesn't exist when creating a user
- [x] When a user account with the role 'student' is deleted, the student_page > student_user relationship is set null
- [ ] For deployment and for testing a  page permission will need to be set up on the site to allow the student group to `edit` the students index page and all pages below it


rca/people/models.py
```diff
@@ -656,6 +662,7 @@ class StudentPage(BasePage):
     ]
 
     content_panels = BasePage.content_panels + [
+        FieldPanel("student_user_account"),
```
richbrennan, line 665: Just to flag that if there are a large number of student accounts the `<select>` could be slow. I'm not sure of the solution, but this might be an option https://github.com/wagtail/wagtail-autocomplete

rca/users/signals.py
```diff
@@ -0,0 +1,43 @@
...
+    user = kwargs["instance"]
+
+    if user.groups.filter(name="Students").exists():
+        try:
+            student_index = StudentIndexPage.objects.first()
+        except StudentIndexPage.DoesNotExist:
+            return
+
+        student_page = StudentPage.objects.filter(
+            first_name=user.first_name, last_name=user.last_name,
```
richbrennan, line 23: Is first and last name a reliable way to check, having more than one student with the same name seems quite feasible?

rca/users/signals.py
```diff
@@ -0,0 +1,43 @@
...
+        )
+        if student_page:
```
richbrennan, line 25: I think you could use exists again:

```
if StudentPage.objects.filter(...).exists():
    logger.info(...)
else:
    student_index.add_child(
            instance=StudentPage(
                first_name=user.first_name,
                last_name=user.last_name,
                title=f"{user.first_name} {user.last_name}",
                student_user_account=user,
                live=False,
            )
        )
```

richbrennan (commented):
A few comments, thanks Kevin

rca/users/signals.py
kevinhowbrook: Yea I'm not honestly too sure, we do have emails but the possibility of a different email between user and page I think is higher than names changing.

rca/people/models.py
kevinhowbrook: Thanks, that's worth noting, I'll add it to the ticket

rca/users/signals.py
kevinhowbrook: Up coming in #618 I'm adding a field to 'create_student_page' so this adds an extra measure

PR #624, opened 2021-04-19
728 student profile creation (student journey)

kevinhowbrook:
Ticket [728](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/728)

In summary, I need a way for administrators to add a user account with the role 'student', without adding a password. This will be so I can interrupt the account creation, add a password myself and send out a reset password notification email. I couldn't quite see a way of doing that on the rca.users.model. So I've created a view and form to add a 'student'. 

TODO 
- [x] Change base of merge to p3-student-profile once #618 is merged
- [x] Shift signal to form 
- [x] Email notification password reset
- [x] admin UI button for creating students
- [x] permissions on the form
- [x] Validation for adding duplicate users
- [x] Validation for adding usernames with numbers and spaces
- [x] Unit tests for creation of students and student pages
- [x] Redirect to student page after reset password
- [x] Unit tests for students only editing their own page
- [x] Fix csrf issue on student login and redirect

image


rca/account_management/views.py
```diff
@@ -0,0 +1,27 @@
...
+    template_name = "account_management/create.html"
+    form_class = StudentCreateForm
+    success_url = "/admin/student/create"  # TODO - better success template
+
+    def form_valid(self, form):
+        data = super().form_valid(form)
+        student_group = Group.objects.get(name="Students")
+        student_user = User.objects.create_user(
+            username=form.cleaned_data["username"],
+            password="123pass",
```
kevinhowbrook, line 20: Obviously swap this for `user.set_unusable_password()`

rca/account_management/views.py
```diff
@@ -0,0 +1,27 @@
...
+            email=form.cleaned_data["email"],
+            first_name=form.cleaned_data["first_name"],
+            last_name=form.cleaned_data["last_name"],
+        )
+        student_user.groups.add(student_group)
```
kevinhowbrook, line 25: At the moment, as part of the previous ticket, adding this group here will initiate the code in this signal https://github.com/torchbox/rca-wagtail-2019/blob/728-student-profile-student-journey/rca/users/signals.py

I think it might be worth moving the code in that signal to create page into this `form_valid()` method, along with a manual option to 'create_student_page` on the form?

rca/account_management/forms.py
```diff
@@ -0,0 +1,12 @@
+from django import forms
+from django.contrib.auth.forms import UsernameField
+
+
+class StudentCreateForm(forms.Form):
+    # Personal Details
+    username = UsernameField(
+        widget=forms.TextInput(attrs={"autofocus": True}), required=True
+    )
```
kevinhowbrook, line 9: I've taken this `UsernameField` from django, I'm not entirely sure if I should be adding validation for this somewhere though

rca/account_management/views.py
```diff
@@ -0,0 +1,27 @@
...
+from rca.users.models import User
+
+from .forms import StudentCreateForm
+
+
+class CreateStudentFormView(FormView):
+
+    template_name = "account_management/create.html"
+    form_class = StudentCreateForm
+    success_url = "/admin/student/create"  # TODO - better success template
```
kevinhowbrook, line 13: Probably back to the create view with a message added to the admin

rca/account_management/forms.py
alxbridge: Looking at the [django definition](https://github.com/django/django/blob/stable/2.2.x/django/contrib/auth/forms.py#L69) for this field, it's just a `CharField` with a little bit of unicode normalisation.

The corresponding field on AbstractUser is validated with [`UnicodeUsernameValidator`](https://github.com/django/django/blob/7f1b088ab4a4342a87a11496096471703994a006/django/contrib/auth/validators.py#L19) - so either we'll be OK if the validation from AbstractUser carries over to what we're doing, or we might want to also make use of that validator.

rca/account_management/views.py
alxbridge: Sounds good, especially if they might be creating users in a batch

rca/account_management/views.py
alxbridge: If this is going to be the only mechanism by which student users are created, I'd be inclined to agree that this `form_valid()` method would be a good place for the code in the signal.

> along with a manual option to 'create_student_page` on the form?

Can you explain a bit more what you mean here? As in, there'd be a field on the form which could control whether a student page was created? Or something else (e.g. a supplementary form for adding a student page to an existing user)?

kevinhowbrook: Yea my thoughts here are to implement a simple boolean on the form to give more control over whether a page should be created

rca/account_management/forms.py
kevinhowbrook: Switching to a model form has allowed me to make use of the validation here

rca/account_management/views.py
```diff
@@ -0,0 +1,139 @@
...
+                    title=f"{form.cleaned_data['first_name']} {form.cleaned_data['last_name']}",
+                    student_user_account=student_user,
+                    live=False,
+                )
+            )
+
+            # Generate a password reset link to be emailed to the user.
+            password_reset_url = self.request.build_absolute_uri(
+                get_set_password_url(student_user)
+            )
```
kevinhowbrook, line 115: This builds me out the correct URL to send to the user, however I'm struggling to see how to work a redirect to edit the studentpage. I can't do this with a param after the reset link as the user is asked to login using their new password.

So I think I might need to be looking at trying to hook into the login view, and if the user has the role 'Student', look for their page and redirect them. But I'm not sure where to start with this :/

rca/people/admin_forms.py
```diff
@@ -8,6 +8,19 @@ class StudentPageAdminForm(WagtailAdminPageForm):
...
 
+    def __init__(self, *args, **kwargs):
+        print(args)
+        print(kwargs)
+        super(StudentPageAdminForm, self).__init__(*args, **kwargs)
+        # TODO work in request.user.group IS/HAS Student and lock fields
+        readonly_fields = [
+            "first_name",
+            "last_name",
+        ]
```
kevinhowbrook, line 19: This is blocking the last AC in the task. I need to work in a way of getting the user requesting the form here
so I can conditionally lock the fields depending on the users role.

rca/account_management/views.py
alxbridge: I agree that overriding [`LoginView`](https://github.com/wagtail/wagtail/blob/dede38be65f2b61726b99ff935a154e91662fc35/wagtail/admin/views/account.py#L289) would be a good way of going about this - specifically with a custom `get_success_url`. I think for this approach you could do the following:

1. Subclass `LoginView` and customise `get_success_url` as required
2. Override the `wagtailcore_login` path in the site's _urls.py_ to use your new view, instead of this one declared in `wagtail.core.urls`:

```
    path(
        '_util/login/',
        auth_views.LoginView.as_view(template_name=WAGTAIL_FRONTEND_LOGIN_TEMPLATE),
        name='wagtailcore_login'),
```

(If for any reason you're not able to/didn't want to override that path directly, you can set up a new path for the custom view, and set `WAGTAIL_FRONTEND_LOGIN_URL` to use it instead.)

Alternatively, you could look at overriding `PasswordResetCompleteView` somehow, so that the link to log in contains an appropriate parameter for redirecting, but I think I'd prefer the first approach.

rca/people/admin_forms.py
alxbridge: Some of the answers on [this Stackoverflow post](https://stackoverflow.com/questions/2864955/django-how-to-get-current-user-in-admin-forms) might help you here

kevinhowbrook: Marking this as resolved, we are actually tackling this in the next ticket, I'm getting ahead of myself

rca/urls.py
```diff
@@ -10,11 +10,17 @@
...
 from rca.utils.cache import get_default_cache_control_decorator
 from rca.wagtailapi.api import api_router
 
+WAGTAIL_FRONTEND_LOGIN_TEMPLATE = getattr(
+    settings, "WAGTAIL_FRONTEND_LOGIN_TEMPLATE", "wagtailcore/login.html"
+)
 # Private URLs are not meant to be cached.
 private_urlpatterns = [
+    path("admin/login/", CustomLoginView.as_view(), name="wagtailcore_login"),
+    path("admin/_util/login/", CustomLoginView.as_view(), name="wagtailcore_login"),
```
kevinhowbrook, line 23: I have no idea why I need to define both of these, but I do.

rca/account_management/views.py
kevinhowbrook: Thanks @alxbridge, I think I've found a way to override the LoginView here, so I'll get this code reviewed

rca/account_management/views.py
```diff
@@ -0,0 +1,173 @@
...
+            # Notify the user of the account.
```
kevinhowbrook, line 118: AC: When a new student account is created an email notification is sent to the user asking them to reset their password

rca/account_management/views.py
```diff
@@ -0,0 +1,173 @@
...
+                f"A Notification email has been sent to {student_user.email}",
+            )
+
+        return data
+
+
+class CustomLoginView(LoginView):
+    """Custom login view to redirect students to their profile page
+    """
+
```
kevinhowbrook, line 147: AC: As a student after resetting my password, then I can log in and I am redirected to my profile in Wagtail CMS

rca/users/tests/test_models.py
```diff
@@ -1,77 +1,8 @@
...
-            first_name="Monty",
-            last_name="Python",
-        )
-        student_user.groups.add(student_group)
-        self.assertEqual(len(StudentPage.objects.all()), 1)
-        student_page = StudentPage.objects.get(title="Monty Python")
-        self.assertEqual(student_page.student_user_account, student_user)
-        student_user.delete()
-        student_page.refresh_from_db()
-        self.assertEqual(student_page.student_user_account, None)
```
kevinhowbrook, line 77: We are doing account creation in the new account_management app so I've shifted testing the there

rca/urls.py
richbrennan: Strange that they're both allowed to have the same `name="wagtailcore_login"`, maybe one is used by Django itself and the other by Wagtail...?

richbrennan (commented):
A few comments, thanks Kevin!

rca/account_management/tests/test_utils.py
```diff
@@ -0,0 +1,11 @@
...
+
+from rca.account_management.utils import get_set_password_url
+from rca.users.factories import UserFactory
+
+
+class TestAccountManagementUtils(TestCase):
+    def test_set_password_reset_link(self):
+        user = UserFactory()
+        reset_link = get_set_password_url(user)
+        self.assertTrue(reset_link)
```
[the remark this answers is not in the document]
kevinhowbrook: This is pretty hard for me, the str return is different every time as it contains tokens

`/admin/password_reset/confirm/MTEw/alx9ma-fe0dbf17415b85ecb3b934bfff1073cf/`

Do you think I could assert it with a regex pattern?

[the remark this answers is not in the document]
richbrennan: Yes, the Python `TestCase` has an `assertRegex()` method.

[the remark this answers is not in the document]
kevinhowbrook: Added in https://github.com/torchbox/rca-wagtail-2019/pull/624/commits/8b90558bd6ee5a5a007e4196a37a87de83bfd04a

kevinhowbrook:
@richbrennan I think there is just one remaining thing to check here. This new commit https://github.com/torchbox/rca-wagtail-2019/pull/624/commits/a0bab67c677f27781f3adb8325e2d816b1b4cf05

I'm having to publish the created page, then set live=False so the FK relationships are created.

PR #636, opened 2021-04-29
732 Enquiry Form Amend delete submissions

Morsey187:
Ticket: https://projects.torchbox.com/projects/rca-website-rebuild/tickets/732

Adds new date filters to only delete submissions older than 7 days, as well as updating the template text and new UI state where their are submissions but no submissions over 7 days to delete (will confirm the text for that one with Frank).

### DIFF
--- a/rca/enquire_to_study/views.py
+++ b/rca/enquire_to_study/views.py
@@ -322,7 +322,7 @@ def delete(request):
 
     time_threshold = timezone.now() - timedelta(days=7)
     instances = EnquiryFormSubmission.objects.filter(
-        Q(submission_date__gte=time_threshold)
+        Q(submission_date__lte=time_threshold)
     )
     count_delete_submissions = len(instances)
 
--- a/rca/enquire_to_study/views.py
+++ b/rca/enquire_to_study/views.py
@@ -320,10 +320,10 @@ def delete(request):
     url_helper = EnquiryFormSubmissionAdmin().url_helper
     index_url = url_helper.get_action_url("index")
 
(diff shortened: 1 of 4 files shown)


rca/enquire_to_study/views.py
```diff
@@ -218,13 +222,16 @@ def delete(request):
...
+    time_threshold = timezone.now() - timedelta(days=7)
+    instances = EnquiryFormSubmission.objects.filter(
+        Q(submission_date__gte=time_threshold)
+    )
+    count_delete_submissions = len(instances)
+
     if request.method == "POST":
-        for instance in instances:
-            instance.delete()
+        instances.delete()
```
Morsey187, line 232: This should keep the number of queries down

rca/enquire_to_study/views.py
```diff
@@ -233,7 +240,8 @@ def delete(request):
         request,
         "enquire_to_study/confirm_delete.html",
         {
-            "count": count,
+            "count_all_submissions": EnquiryFormSubmission.objects.count(),
```
kevinhowbrook, line 243: I think instead of making 2 queries for `EnquiryFormSubmission.objects` we could do the following

```
count_submissions = EnquiryFormSubmission.objects.all()
instances = count_submissions.objects.filter(
        Q(submission_date__lte=time_threshold)
    )
```

Then use those vars in the context method?

kevinhowbrook (approved):
Just a small suggestion, otherwise looks good to me. Send it my once you add tests for this :)

PR #637, opened 2021-04-30
731 Add CustomEmbedFinder to fix title attribute

Morsey187:
Ticket: https://projects.torchbox.com/projects/rca-website-rebuild/tickets/731

The added tests for the fix passes fine although I haven't been able to manually test on my local build due to this error:
`django.db.utils.IntegrityError: null value in column "hash" of relation "wagtailembeds_embed" violates not-null constraint`
More info on that error here https://github.com/wagtail/wagtail/issues/6764. 
Not sure if that's a problem with data pulled from develop but feel like this should be tackled in a seperate ticket for time constraints.


kevinhowbrook (approved):
Thanks Ben, this looks good to me. Please can we merge it to dev to dest it there please.

PR #642, opened 2021-05-12
728 edit page permissions

kevinhowbrook:
hacky solution to raising permission denied when editing the page, basically because this is the only point I have access to the request at


richbrennan:
Would this hook help avoid a hacky solution? https://docs.wagtail.io/en/v2.12.4/reference/hooks.html#before-edit-page

kevinhowbrook:
Closing this - the adjustments we are making on https://projects.torchbox.com/projects/rca-website-rebuild/tickets/762 will resolve it

PR #643, opened 2021-05-14
762 Student user - specific permissions by groups

kevinhowbrook:
Ticket [762](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/762)

@richbrennan  this is a bit of a change-up so it might be worth chatting about

- [x] Remove the global images permission on the student group 
- [x] Can we make it so that the student cannot change their username - as this has side effects
- [x] New unit test for the process, saving student page creates group with permissions etc
- [x] Lock down user and image fields once set 

The following are being pulled into a new [ticket](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/776):

- [ ] change OTO fields to use on_delete.PROTECT
- [ ] Or, pre delete signal on collection and user to check


rca/account_management/views.py
```diff
@@ -119,6 +125,37 @@ def form_valid(self, form):
             student_page.save_revision()
             student_page.unpublish()
 
+            # Create a specific group for this student so they have edit access to their page
+            # and their image collection
+            specific_student_group, created = Group.objects.get_or_create(
+                name=f"Student: {form.cleaned_data['first_name']} {form.cleaned_data['last_name']}"
```
kevinhowbrook, line 131: Todo - change naming convention to username as this will be unique

rca/people/models.py
```diff
@@ -872,6 +880,44 @@ def student_related_links(self):
             for item in self.relatedlinks.all()
         ]
 
+    def save(self, *args, **kwargs):
+        """On saving the student page, make sure the student_user_account
+        has a group created with the necessary permissions
+        """
+        super(StudentPage, self).save()
+        if self.student_user_image_collection and self.student_user_image_collection:
```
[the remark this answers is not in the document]
kevinhowbrook: Also we should add validation for _both_ these fields being required if one is set.
There won't be a case where we need only one of these values

rca/account_management/views.py
```diff
@@ -119,6 +125,37 @@ def form_valid(self, form):
...
+                ),
+                permission=Permission.objects.get(codename="add_image"),
+            )
+            GroupCollectionPermission.objects.get_or_create(
+                group=specific_student_group,
+                collection=Collection.objects.get(
+                    name=form.cleaned_data["student_user_image_collection"]
+                ),
+                permission=Permission.objects.get(codename="choose_image"),
+            )
```
kevinhowbrook, line 153: Pull this out to a method in utils so code isn't duplicated in the save student_page method

rca/people/models.py
```diff
@@ -872,6 +880,44 @@ def student_related_links(self):
...
+                    collection=Collection.objects.get(
+                        name=self.student_user_image_collection
+                    ),
+                    permission=Permission.objects.get(codename="choose_image"),
+                )
+
+            # Add the new specific student group to the user
+            self.student_user_account.groups.add(specific_student_group)
+            self.student_user_account.save()
+
```
kevinhowbrook, line 920: This needs covering by a unit test

rca/people/models.py
```diff
@@ -872,6 +880,44 @@ def student_related_links(self):
...
+                    group=specific_student_group,
+                    collection=Collection.objects.get(
+                        name=self.student_user_image_collection
+                    ),
+                    permission=Permission.objects.get(codename="choose_image"),
+                )
+
+            # Add the new specific student group to the user
+            self.student_user_account.groups.add(specific_student_group)
+            self.student_user_account.save()
```
kevinhowbrook, line 919: This should be indented to work under the `if created` condition

rca/account_management/views.py
kevinhowbrook, line 153: remove `get` here - it's masking potential bugs - this journey would always be creating new groups

rca/people/models.py
```diff
@@ -872,6 +880,44 @@ def student_related_links(self):
...
 
+    def save(self, *args, **kwargs):
+        """On saving the student page, make sure the student_user_account
+        has a group created with the necessary permissions
+        """
+        super(StudentPage, self).save()
+        if self.student_user_image_collection and self.student_user_image_collection:
+            # Create a specific group for this student so they have edit access to their page
+            # and their image collection
+            specific_student_group, created = Group.objects.get_or_create(
```
kevinhowbrook, line 891: We should check for the `GroupPagePermission` existing first before creating anything.

rca/people/models.py
```diff
@@ -872,6 +880,44 @@ def student_related_links(self):
             for item in self.relatedlinks.all()
         ]
 
+    def save(self, *args, **kwargs):
+        """On saving the student page, make sure the student_user_account
+        has a group created with the necessary permissions
+        """
```
kevinhowbrook, line 886: To do - only create these permissions if:
- New self.id: the page is new
- image_collection or student_user values on the page have changed

kevinhowbrook: Potentially move this to the forms save method too https://docs.wagtail.io/en/stable/advanced_topics/customisation/page_editing_interface.html#customising-generated-forms

rca/account_management/views.py
kevinhowbrook: This is now in one place in the `StudentPage.save()`

rca/people/models.py
kevinhowbrook: This isn't going to happen as part of this task. Due to all the custom form things and so many moving parts I'm not able to even test form.valid() at the moment.

rca/people/admin_forms.py
```diff
@@ -5,4 +5,23 @@ class StudentPageAdminForm(WagtailAdminPageForm):
...
+                "a student user account must be added.",
+            )
+        if student_user_account and not student_user_image_collection:
+            self.add_error(
+                "student_user_image_collection",
+                "If you are adding a student user account so a student can access "
+                "this page, an image collection must be added.",
+            )
+
+        return cleaned_data
```
kevinhowbrook, line 57: for testing this, check we can load the form via `page.edit_handler.get_form`

rca/people/models.py
kevinhowbrook: Is not creating these relationships when moving this to a form down to commit=False ?

rca/people/admin_forms.py
```diff
@@ -5,4 +5,53 @@ class StudentPageAdminForm(WagtailAdminPageForm):
     """Validation for adding StudentPage
     """
 
-    pass
+    def __init__(self, *args, **kwargs):
+        super(StudentPageAdminForm, self).__init__(*args, **kwargs)
+        # Don't allow changing the user account or image collection once set as the are tied to permissions
```
kevinhowbrook, line 10: typo - they

rca/people/admin_forms.py
```diff
@@ -5,4 +5,53 @@ class StudentPageAdminForm(WagtailAdminPageForm):
...
+                except KeyError:
+                    pass
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        instance = getattr(self, "instance", None)
+
+        # # As these fields are disabled once set, they need adding back into the post data
+        if instance.student_user_image_collection and instance.student_user_account:
```
kevinhowbrook, line 33: This will throw an Attribute error, check for `if instance and...` too.

rca/users/forms.py
```diff
@@ -0,0 +1,8 @@
+from wagtail.users.forms import UserEditForm
+
+
+class CustomUserEditForm(UserEditForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # TODO This also needs to be removed from the POST data
```
kevinhowbrook, line 7: remove todo

rca/people/models.py
kevinhowbrook: TODO - finish tests for creating page with groups

rca/people/admin_forms.py
kevinhowbrook: Still no dice with that

rca/people/tests/test_forms.py
```diff
@@ -0,0 +1,96 @@
...
+            }
+        )
+
+    def get_form(self, instance, data=None):
+        return self.form_class(instance=instance, data=data)
+
+    def test_form_not_valid(self):
+        print(self.get_form_data())
+        form = self.get_form(instance=self.student_page, data=self.get_form_data())
+        self.assertFalse(form.is_valid())
```
kevinhowbrook, line 96: This isn't passing, the form.is_valid() here should be false, because we aren't setting anything as the student_user_account on line [86](https://github.com/torchbox/rca-wagtail-2019/pull/647/files#diff-8ccc24b43f569e4fc9146e3d79c9e6d2f0e55dfff11c6b60c755d0838769478fR86). It's testing the validation is working [here](https://github.com/torchbox/rca-wagtail-2019/blob/471be5c4d5eca1d1bd5ed2ba81b9304263bff4cf/rca/people/admin_forms.py#L44-L54). In the admin UI this is fine.

rca/people/tests/test_models.py
```diff
@@ -28,3 +30,39 @@ def test_page_count_rules(self):
...
+                title="Dana Scully",
+                slug="a-student",
+                first_name="dana",
+                last_name="student",
+                student_user_image_collection=self.collection,
+                student_user_account=self.student
+                # {'student_user_account': ['user instance with id 3 does not exist.']}
+                # I do not understand why, I can see it exists!
+            )
+        )
```
kevinhowbrook, line 67: As the comment states here, I've no idea why the user isn't available in the test run.
In the debug here https://github.com/torchbox/rca-wagtail-2019/blob/60a31e25a0348b3729d3c71c9da183b3bb055763/rca/people/models.py#L887 on saving the page via the admin, I see the user in the logs, but in the test run this prints None

PR #647, opened 2021-05-24
762 tests

kevinhowbrook:


rca/people/tests/test_forms.py
```diff
@@ -0,0 +1,96 @@
...
+            }
+        )
+
+    def get_form(self, instance, data=None):
+        return self.form_class(instance=instance, data=data)
+
+    def test_form_not_valid(self):
+        print(self.get_form_data())
+        form = self.get_form(instance=self.student_page, data=self.get_form_data())
+        self.assertFalse(form.is_valid())
```
kevinhowbrook, line 96: This isn't passing, the form.is_valid() here should be false, because we aren't setting anything as the student_user_account on line [86](https://github.com/torchbox/rca-wagtail-2019/pull/647/files#diff-8ccc24b43f569e4fc9146e3d79c9e6d2f0e55dfff11c6b60c755d0838769478fR86). It's testing the validation is working [here](https://github.com/torchbox/rca-wagtail-2019/blob/471be5c4d5eca1d1bd5ed2ba81b9304263bff4cf/rca/people/admin_forms.py#L44-L54). In the admin UI this is fine.

rca/people/tests/test_models.py
```diff
@@ -28,3 +30,39 @@ def test_page_count_rules(self):
...
+        self.student_index.add_child(
+            instance=StudentPage(
+                title="Dana Scully",
+                slug="a-student",
+                first_name="dana",
+                last_name="student",
+                student_user_image_collection=self.collection,
+                student_user_account=self.student
+                # {'student_user_account': ['user instance with id 3 does not exist.']}
+                # I do not understand why, I can see it exists!
```
kevinhowbrook, line 65: As the comment states here, I've no idea why the user isn't available in the test run.
In the debug here https://github.com/torchbox/rca-wagtail-2019/blob/60a31e25a0348b3729d3c71c9da183b3bb055763/rca/people/models.py#L887 on saving the page via the admin, I see the user in the logs, but in the test run this prints None

PR #657, opened 2021-06-30
789 Editorial Page - Base and key details

larathompson:
Ticket: https://projects.torchbox.com/projects/rca-website-rebuild/tickets/789


rca/editorial/models.py
```diff
@@ -0,0 +1,116 @@
...
+
+class EditorialPageRelatedSchoolsAndResearchPages(RelatedPage):
+    source_page = ParentalKey(
+        "EditorialPage", related_name="related_schools_and_research_pages"
+    )
+    panels = [
+        PageChooserPanel("page", ["schools.SchoolPage", "research.ResearchCentrePage"])
+    ]
+
+    api_fields = [APIField("page")]
```
larathompson, line 24: Do I need this line? It was used in the ProgrammePage but not sure if it is needed here. From looking online, it looks like this field defines the fields that will be accessible in the API, just not entirely sure what the difference is in this case with/without the field..

rca/editorial/models.py
```diff
@@ -0,0 +1,116 @@
...
+class Author(models.Model):
+    name = models.CharField(max_length=128)
+
+    def __str__(self):
+        return self.name
+
+
+class EditorialPageAuthor(models.Model):
+    page = ParentalKey("EditorialPage", related_name="editorial_authors")
+    author = models.ForeignKey("Author", related_name="+", on_delete=models.CASCADE)
```
larathompson, line 36: What does related_name="+" mean. I see it used q. a lot just not sure why the '+' is always used

rca/editorial/models.py
```diff
@@ -0,0 +1,116 @@
...
+    panels = [FieldPanel("author")]
+
+    def __str__(self):
+        return self.author.name
+
+
+class EditorialPage(BasePage):
+
+    intro_text = RichTextField(blank=True, max_length=255)
```
larathompson, line 45: Maybe the max_length does not need to be defined?

rca/editorial/models.py
```diff
@@ -0,0 +1,116 @@
...
+        "images.CustomImage",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+        related_name="+",
+    )
+
+    #  do you want it to be from date created - auto_now_add (I think )
+    #  or from date_updated = add_add
+    date = models.DateTimeField(auto_now_add=True)
```
larathompson, line 70: I think we want auto_add_now as this adds a date which relates to when the object was first created eg. when the article was first published?

rca/editorial/models.py
kevinhowbrook: I'd remove it - we may end up needing it - but for the purposes of this ticket we aren't interested in defining this right now

rca/editorial/models.py
kevinhowbrook: The + is a way of declaring the relationship won't work backwards.

> If you’d prefer Django not to create a backwards relation, set related_name to '+' or end it with '+'. For example, this will ensure that the User model won’t have a backwards relation to this model:

https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.ForeignKey.related_name

rca/editorial/models.py
kevinhowbrook: Hm - looking at the designs I think the max length makes sense for the introduction field. But this should just be a text field, it doesn't need to be rich text as far as I know

rca/editorial/models.py
kevinhowbrook: :+1:

rca/editorial/models.py
```diff
@@ -0,0 +1,116 @@
...
+    PageChooserPanel,
+)
+from wagtail.api import APIField
+from wagtail.core.fields import RichTextField
+from wagtail.images.edit_handlers import ImageChooserPanel
+
+from rca.utils.models import BasePage, RelatedPage
+
+
+class EditorialPageRelatedSchoolsAndResearchPages(RelatedPage):
```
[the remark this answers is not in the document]
kevinhowbrook: Both is correct, in a code sense, SchoolPage and ResearchCentrePage are both page models. I'm sure I understand the question fully here though, are you asking about the RelatedPage inheritance?

kevinhowbrook:
Note: We'll want to change the merge target here to the integration branch which will be `p3-2-sprint-1`, I'll check in with nick on where this is up to

rca/editorial/models.py
```diff
@@ -0,0 +1,116 @@
...
+    #     taxonomy_tags = []
+    #     if self.related_school_pages:
+    #         for i in self.related_school_pages.all():
+    #             taxonomy_tags.append({"title": i.page.title})
+    #     context["taxonomy_tags"] = taxonomy_tags
+
+    #     return context
+
+    # have not included date here as I tink it should be added automiatically
+    content_panels = BasePage.content_panels + [
```
kevinhowbrook, line 99: Missing 'Data; field from the content panels

rca/editorial/models.py
```diff
@@ -0,0 +1,116 @@
...
+class EditorialPageAuthor(models.Model):
+    page = ParentalKey("EditorialPage", related_name="editorial_authors")
+    author = models.ForeignKey("Author", related_name="+", on_delete=models.CASCADE)
+    panels = [FieldPanel("author")]
+
+    def __str__(self):
+        return self.author.name
+
+
+class EditorialPage(BasePage):
```
kevinhowbrook, line 43: I think we are missing a field for the following acceptance:

> As a publisher I can relate add an 'area' taxonomy value

rca/editorial/models.py
```diff
@@ -0,0 +1,116 @@
...
+            ],
+            heading="Introductory Video",
+        ),
+        FieldPanel("email"),
+        MultiFieldPanel(
+            [InlinePanel("related_schools_and_research_pages")],
+            heading="Related Schools and Research Centres",
+        ),
+        InlinePanel("editorial_authors", label="Editorial Authors"),
+    ]
```
kevinhowbrook, line 116: **Request**

Some of the panels here should be in a new tab 'Key Details' in the admin area (example on the project page [https://rca-development.herokuapp.com/admin/pages/558/edit/]))

We define them like this https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/projects/models.py#L227

And configure them like so: https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/projects/models.py#L246-L253

rca/editorial/models.py
larathompson: Changed to `intro_text = models.CharField(blank=True, max_length=255)`

rca/editorial/models.py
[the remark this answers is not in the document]
larathompson: You explained yesterday but i was confused before as I did not realise that the Schools & Centres were pages (not entirely sure what I thought they were before!)

rca/editorial/models.py
larathompson: Do I want to add the date field? Will this not be added automatically from when it's published?

larathompson: If I add `auto_now_add` to the `datefield`, I think it makes the date non-editable and so it can't be added to the content panels. Unless, we don't use `auto_now_add` and the publisher can select a date for the Editorial Page?

kevinhowbrook: Hadn't realized auto_add_now would do that. Our AC states:

> As a publisher I can add a Date value to the Editorial page

So the main thing we need here is the ability to set an article date. We should also make this a required field too as it's what we will be automatically listing articles by. Maybe don't worry about auto_add_now. Strictly speaking, it's not a requirement but a nice to have

larathompson:
Screenshot 2021-07-09 at 11 05 47

larathompson:
Screenshot 2021-07-09 at 11 28 15

kevinhowbrook:
Thanks @larathompson, could we please un-link the taxonomy tags on https://github.com/torchbox/rca-wagtail-2019/blob/bf03f065d656cd873b7930ca2396de4ed926e2e1/rca/project_styleguide/templates/patterns/molecules/key-details/key-details--editorial.html#L37

kevinhowbrook (changes requested):
Thanks Lara, I've explained on the ticket but I think the only think left here is to remove the <a> tags on the taxonomy terms https://github.com/torchbox/rca-wagtail-2019/blob/bf03f065d656cd873b7930ca2396de4ed926e2e1/rca/project_styleguide/templates/patterns/molecules/key-details/key-details--editorial.html#L37 then this can be merged to dev for testing

kevinhowbrook (approved):

kevinhowbrook:
rebased on p3.2/sprint-1

PR #664, opened 2021-07-06
WIP: Feature/810 editorial page body

larathompson:
Ticket: https://projects.torchbox.com/projects/rca-website-rebuild/tickets/810

Linked to: https://projects.torchbox.com/projects/rca-website-rebuild/tickets/789

Tasks
- Configure quote carousel block (Gallery and Accordion blocks already exist in utils)
- Configure body field on the Editorial Page as a streamfield to include all required blocks


rca/editorial/models.py
```diff
@@ -0,0 +1,145 @@
...
+                    "related_schools_and_research_pages",
+                    label="School or Research Centre",
+                ),
+                InlinePanel("areas", label="Area"),
+            ],
+            heading="Related School, Research Centre or Area",
+        ),
+        FieldPanel("author"),
+        FieldPanel("contact_email"),
+        StreamFieldPanel("body"),
```
kevinhowbrook, line 119: **Blocking**
I think this needs to be in the content_panels section so it appears in the right section in the back end

larathompson:
This branch has been created from https://github.com/torchbox/rca-wagtail-2019/pull/657/commits - this branch has been merged to dev.

larathompson:
Two questions for the Body page:
- Are there only certain videos that can be embedded? Is this something that will be addressed in another ticket (you mentioned something about issues with Youtube before)
- I don't think that the header is displaying correctly (its much bigger in the designs) - again, is this something I should leave as it will be sorted on the FE on another ticket

larathompson:
Question:
If an 'introduction_image' is excluded on the editorial page, then the 'key details' section appears twice. I have rebased from master and this is still an issue. It is due to this bit of code i think in the `editorial_detail.html`

`  <div class="section__row section__row--first-small section__row--last grid">
                    {% if not page.introduction_image %}
                        <div class="layout__start-one layout__span-two layout__@large-span-one u-large-bp-only">
                            {% include "patterns/molecules/key-details/key-details--editorial.html" %}
                        </div>
                    {% endif %}`

I am not sure if i should raise this on a separate ticket?

Screenshot 2021-07-14 at 14 43 24

kevinhowbrook:
@larathompson let's not be concerned with the heading here and just keep it to the body content. RE videos I think this is more down to youtube than any rules we have set up for embedding.

The duplicate key details is a bit puzzling, I tried it locally and had to refresh the page once removing the video image. 

So all in all I think this is good :+1:

kevinhowbrook (approved):

PR #669, opened 2021-07-07
Add groups to admin user listing

tombola:
https://projects.torchbox.com/projects/rca-django-cms-project/tickets/1208

This was a lot simpler than I thought it would be. Initially I overode the URLs and views for the user listing, and then found I was only changing anything in the listing template anyway. 

It is not particularly performant to query the groups per user but I couldn't work out an easy way to improve this.

### DIFF
--- a/rca/people/models.py
+++ b/rca/people/models.py
@@ -890,7 +890,7 @@ class StudentPage(PerUserPageMixin, BasePage):
 
             # Check if a group configuration exsists already for this user.
             group = Group.objects.filter(
-                name=f"Student: {self.student_user_account.username}"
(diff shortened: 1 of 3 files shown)


kevinhowbrook (commented):
Code looks good to me - I'm failing to see this on the staging site though - or am I missing something?

tombola:
I didn't merge this to staging, but it looked this morning like CI had failed on the deploy staging action, for no reason that I could determine. Re-running the task appeared to deploy successfully, at least in circle CI, but I can't see any build or release on heroku, so I have triggered build again by pushing staging directly to heroku.

PR #672, opened 2021-07-08
Feature/786 large cta

kevinhowbrook:
Ticket [786](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/786)

This implements a [LargeCTA](https://github.com/torchbox/rca-wagtail-2019/pull/672/files#diff-8727cdfddca4af8c1d57767617a04f7b815336bec165334bc10ff85fe8d90c59R724-R739) which will be used on Events, Editorial and the new landing pages for alumni and development. It's identical to the Contact fields mixin we currently have, the only difference being is that we would like different field labels for it as it's not going to be used for contact data.


rca/utils/models.py
```diff
@@ -719,3 +719,21 @@ def clean(self):
             "Contact",
         )
     ]
+
+
+class LargeCTAMixin(ContactFieldsMixin):
```
richbrennan, line 724: Suspect use of inheritance, `LargeCTAMixin` is a `ContactFieldsMixin`...? :thinking: 

They seem like different models that serve a similar, but not identical, purpose, even if they have the same field types. Using inheritance + hack ties them together when no relationship necessarily exists, risking future maintenance issues if one were to change. Any reason not to KIS and have separate models?

richbrennan (changes requested):

rca/utils/models.py
kevinhowbrook: Yea, you're right - there no real need to work the models this way - have re-done this with a model form

richbrennan (approved):

PR #674, opened 2021-07-08
Feature/800 fe editorial listing template

nicklee:

PR #681, opened 2021-07-13
788 - editorial listing

kevinhowbrook:
Contains #674 (FE template work) and #657 (BE editorial page base)

Ticket [788](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/788)


rca/editorial/models.py
```diff
@@ -127,3 +128,100 @@ def get_context(self, request, *args, **kwargs):
...
+                if page.related_schools_and_research_pages.exists():
+                    meta = page.related_schools_and_research_pages.first().page.title
+
+                related_pages.append(
+                    {
+                        "title": page.title,
+                        "link": page.url,
+                        "image": page.listing_image or page.hero_image,
+                        "description": page.introduction
+                        if hasattr(page, "introduction")
```
richbrennan, line 174: This will use `introduction` even if its blank, so `listing_summary` will never be used in preference (in case it needs to be)

richbrennan (changes requested):

rca/editorial/models.py
```diff
@@ -187,10 +195,11 @@ class EditorialListingPage(BasePage):
...
         related_pages = []
-        pages = self.related_editorial_pages.all()
-        for value in pages.select_related("page"):
-            if value.page and value.page.live:
-                page = value.page.specific
+        pages = self.related_editorial_pages.all().select_related("page")
+        pages = pages.prefetch_related("page__hero_image", "page__listing_image")
+        for value in pages:
+            page = value.page
+            if page and page.live:
```
richbrennan, line 202: Could the live check move to the original query, either with `.live()` or `filter(live=True)`?

rca/editorial/models.py
```diff
@@ -187,10 +195,11 @@ class EditorialListingPage(BasePage):
 
     def get_editor_picks(self):
         related_pages = []
-        pages = self.related_editorial_pages.all()
-        for value in pages.select_related("page"):
-            if value.page and value.page.live:
-                page = value.page.specific
+        pages = self.related_editorial_pages.all().select_related("page")
+        pages = pages.prefetch_related("page__hero_image", "page__listing_image")
```
richbrennan, line 199: I think you can consolidate these two lines into one:

`pages = self.related_editorial_pages.all().select_related("page__hero_image",  "page__listing_image")`

or maybe

`pages = self.related_editorial_pages.all().prefetch_related("page__hero_image",  "page__listing_image")`

PR #684, opened 2021-07-13
[875] add event type taxonomy

richbrennan:
Implements: https://projects.torchbox.com/projects/rca-website-rebuild/tickets/875

TODO: add migration when merged to event page feature branch

### DIFF
--- a/rca/events/models.py
+++ b/rca/events/models.py
@@ -11,6 +11,16 @@ from rca.utils.models import BasePage
 from .blocks import CallToAction, EventDetailPageBlock
 
 
+class EventType(models.Model):
+    title = models.CharField(max_length=100)
+
+    class Meta:
+        ordering = ["title"]
+
+    def __str__(self):
+        return self.title
+
+
 class EventIndexPage(BasePage):
     subpage_types = ["EventDetailPage"]
     template = "patterns/pages/events/event_index_page.html"
@@ -35,6 +45,12 @@ class EventDetailPage(BasePage):
         help_text="Enter the end date of the event. This will be the same as "
         "the start date for single day events."
     )
+    event_type = models.ForeignKey(
+        EventType,
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name="events",
+    )
     introduction = models.TextField()
     body = StreamField(EventDetailPageBlock())
     call_to_action = StreamField(CallToAction(max_num=1, required=False), blank=True)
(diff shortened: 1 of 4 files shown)


rca/events/models.py
```diff
@@ -11,6 +11,16 @@
...
+class EventType(models.Model):
+    title = models.CharField(max_length=100)
+
+    class Meta:
+        ordering = ["title"]
+
+    def __str__(self):
+        return self.title
+
+
```
kevinhowbrook, line 23:
```suggestion
class EventType(models.Model):
    title = models.CharField(max_length=100)
    slug = models.SlugField(blank=True)

    def save(self, *args, **kwargs):
        self.slug = slugify(self.title)
        super(EventType, self).save(*args, **kwargs)
        
    class Meta:
        ordering = ["title"]

    def __str__(self):
        return self.title
```

**Beg**
Could we sluggify this taxonomy as it's going to be used as a filter for the event listing page, and we'll need it.

kevinhowbrook (approved):

rca/events/models.py
richbrennan: Is there a reason not to slugify in the template for filtering, keep the model simpler?

`<option value={{ event_type|slugify` }}>{{ event_type }}</option>`

richbrennan: In the above suggestion, if the user enters a value it gets blatted on save.

richbrennan: >Is there a reason not to slugify in the template for filtering, keep the model simpler?

Or if not in the template, in `get_context` when constructing the values to pass to the template. Either option reduces the redundancy of keeping slug on the model.

kevinhowbrook: Yea, the reason we slugify these on the model is for when the filtering system, it uses this slug value as a URL arg to filter https://github.com/torchbox/rca-wagtail-2019/pull/686/files#diff-0157f045a04308cf0e5bbc6ce0256711ba1d8bc466a5b54ec6e9c24818e0e39bR274
```
filter_by="subjects__subject__slug__in",
                option_value_field="slug",
```
Feel free to pass on this here, we can pick it up as part of https://projects.torchbox.com/projects/rca-website-rebuild/tickets/787 where we configure the filters

richbrennan: Ugh, OK not our usual method of filtering then. Might as well do it here.

PR #686, opened 2021-07-14
Feature/807 editorial listing filters

kevinhowbrook:
**Contains #681**

Implements listing page filters for https://projects.torchbox.com/projects/rca-website-rebuild/tickets/807

This PR also re-works out the AreaOfExpertise taxonomy and replaces it with Directorate as per [this comment](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/807#update-67716821)

I've also split the related schools and research pages out so the filter system is consistent to where it's used in projects etc


rca/editorial/models.py
```diff
@@ -123,14 +158,16 @@ class EditorialPage(ContactFieldsMixin, BasePage):
         FieldPanel("published_at"),
         MultiFieldPanel(
             [
+                InlinePanel("related_schools", label="Related Schools"),
```
kevinhowbrook, line 161: switched to use the already exposed related school page on https://github.com/torchbox/rca-wagtail-2019/blob/master/rca/schools/models.py#L51

rca/editorial/utils.py
```diff
@@ -0,0 +1,66 @@
...
+    If the EditorialPage we are viewing has a parent listing page then we
+    can link the taxonomy items to the parent listing page and apply them
+    as filter values.
+
+    Since an editorial page can be placed anywhere in the site, this won't
+    always be the case so we need to allow these taxonomy terms to not link.
+
+    Args:
+        page :EditorialPage object
+        parent: The EditorialPage parent object
```
richbrennan, line 14: `parent` isn't in the function signature?

rca/programmes/migrations/0062_add_subject_slugs.py
```diff
@@ -0,0 +1,21 @@
+from django.db import migrations, models
+
+from rca.programmes.models import Subject
```
richbrennan, line 3: models should be imported using `apps.get_model()` https://docs.djangoproject.com/en/3.2/topics/migrations/#data-migrations

rca/settings/base.py
```diff
@@ -763,3 +763,7 @@
 
 pixel_limit = env.get("WAGTAILIMAGES_MAX_IMAGE_PIXELS")
 WAGTAILIMAGES_MAX_IMAGE_PIXELS = int(pixel_limit) if pixel_limit else 10000000
+
+ALLOW_EDITORIAL_PAGE_GENERATION = (
+    env.get("ALLOW_EDITORIAL_PAGE_GENERATION", "true").lower() == "true"
```
richbrennan, line 768: This will default to `True`, would if be safer to default to `False`?

`env.get("ALLOW_EDITORIAL_PAGE_GENERATION", "false")`

richbrennan (commented):
A few comments, thanks @kevinhowbrook

PR #688, opened 2021-07-15
763 ambassador platform

nickmoreton:
## Implementing The Ambassador Platform for RCA 
These changes are the outcome of estimate ticket: 
https://projects.torchbox.com/projects/rca-website-rebuild/tickets/763

I don't think we will need any specific FE work but It's not been possible to use the Actual TAP Carousel or Widgets in place during development. As the ambassador platform admin need to know the domain we will be using we may need to test the appearance on a 'dev/staging' site. As far as I could tell it's not possible to test on localhost.


rca/home/models.py
```diff
@@ -259,6 +266,8 @@ def get_context(self, request, *args, **kwargs):
         context["news_and_events"] = get_news_and_events()
         context["alumni_stories"] = get_alumni_stories()
         context["hero_colour"] = LIGHT_HERO
+        context["tap_widget_code"] = mark_safe(self.tap_widget.script_code)
```
kevinhowbrook, line 269: **Issue**
I think this will throw an attribute error if no tap widget is set, as tap_widget here could be a NoneType if not set

rca/landingpages/models.py
```diff
@@ -549,12 +560,68 @@ def get_context(self, request, *args, **kwargs):
         context = super().get_context(request, *args, **kwargs)
         context["featured_image_secondary"] = self.get_featured_image_secondary()
         context["page_list"] = self.get_page_list()
-
+        if self.tap_widget:
+            context["tap_widget_code"] = mark_safe(self.tap_widget.script_code)
```
kevinhowbrook, line 564: **Issue**
I think this will throw an attribute error if no tap widget is set, as tap_widget here could be a NoneType if not set

kevinhowbrook (changes requested):
Thanks @nickmoreton just a few fixes needed,

As a suggestion, I think we could reduce some of the duplicate code here by swapping the implementation of the FK to the snippet to use an abstract class?

EG:

```
class TapMixin(models.Model):
   tap_widget = models.ForeignKey(
        "utils.TapWidgetSnippet", on_delete=models.SET_NULL, null=True, blank=True,
    )
    
    panels = [SnippetChooserPanel("tap_widget")]


class APageModel(TapMixin, BasePage):
    ...etc

```

nickmoreton:
Thanks @kevinhowbrook for the review. Agreed with both just looking now.

rca/home/models.py
nickmoreton: Agreed I missed that one. I have added the if into place

nickmoreton:
> Thanks @nickmoreton just a few fixes needed,
> 
> As a suggestion, I think we could reduce some of the duplicate code here by swapping the implementation of the FK to the snippet to use an abstract class?
> 
> EG:
> 
> ```
> class TapMixin(models.Model):
>    tap_widget = models.ForeignKey(
>         "utils.TapWidgetSnippet", on_delete=models.SET_NULL, null=True, blank=True,
>     )
>     
>     panels = [SnippetChooserPanel("tap_widget")]
>     
>     class Meta:
>         abstract = True
> 
> 
> class APageModel(TapMixin, BasePage):
>     ...etc
> ```

Also moved the tap_widget field into the TapMixin

nickmoreton:
Hey @kevinhowbrook changes are in place now. Many thanks.

kevinhowbrook (approved):

kevinhowbrook:
All rebased on master

PR #698, opened 2021-07-22
Feature/907 ee news

kevinhowbrook:
Ticket [907](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/907)

Implements the News component on the EE Landing Page


rca/landingpages/models.py
```diff
@@ -560,9 +562,61 @@ class Meta:
         verbose_name = "Landing Page - Enterprise"
 
 
+class EELandingPageRelatedEditorialPage(RelatedPage):
+    source_page = ParentalKey(
+        "landingpages.EELandingPage", related_name="related_editorial_pages"
+    )
+    panels = [PageChooserPanel("page", ["editorial.EditorialPage"])]
```
richbrennan, line 569: **Question** Would it be worth being specific with the page type to aid later queries?

rca/landingpages/models.py
```diff
@@ -560,9 +562,61 @@ class Meta:
...
+
+
 class EELandingPage(BasePage):
     template = "patterns/pages/editorial_event_landing/editorial_event_landing.html"
     max_count = 1
 
+    news_link_text = models.TextField(
+        max_length=120,
+        blank=False,
+        null=True,
```
richbrennan, line 579: `null=True` not required for text field

rca/landingpages/models.py
```diff
@@ -560,9 +562,61 @@ class Meta:
...
+        help_text=_("The text do display for the link"),
+    )
+    news_link_target_url = models.URLField(blank=False, null=True)
```
richbrennan, line 582: `null=True` not required (text field)

rca/landingpages/models.py
```diff
@@ -560,9 +562,61 @@ class Meta:
...
+
     class Meta:
         verbose_name = "Landing Page - Editorial and Events"
+
+    @property
+    def news(self):
```
richbrennan, line 588: Maybe either make this a cached property or make it a method and call in get page context (I think I favour the latter these days, I read somewhere (vague) its best not have DB queries hidden behind properties)

rca/landingpages/models.py
```diff
@@ -560,9 +562,61 @@ class Meta:
...
+        news = []
+        picked_news = self.related_editorial_pages.first().page.specific
```
richbrennan, line 590: This will throw a value error if `related_editorial_pages` was ever empty. I know its required but the linked page could be deleted (possibly an edge case, but if it can happen it probably will!)

rca/landingpages/utils.py
```diff
@@ -0,0 +1,14 @@
+def news_teaser_formatter(item, image=None):
+    item_as_dict = {}
+
+    editorial_type = item.editorial_types.first()
+    item_as_dict["type"] = editorial_type.type if editorial_type else ""
+
+    if image and item.hero_image:
+        item_as_dict["image"] = item.hero_image.file.url
```
richbrennan, line 8: **Question** I think this will return the URL for the original image, does it need to use a particular rendition?

richbrennan (changes requested):
A few comments, thanks Kevin!

rca/landingpages/models.py
kevinhowbrook: I'm not sure what you mean by this?

rca/landingpages/models.py
kevinhowbrook: Swapped for a method and called in get_context

rca/landingpages/models.py
richbrennan: `RelatedPage.page` is a FK to `wagtailcore.Page`, if these pages will always be `EditorialPage`, might it be worth having a FK to `EditorialPage` instead, to make queries easier?

(feel free to ignore this comment :-) )

PR #724, opened 2021-07-27
Feature/908 ee events

kevinhowbrook:
Addresses feedback on ticket [908](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/908#update-67843195)

@richbrennan I'm unsure if this is the best way to deal with this, but we could benefit from a single function for getting the right listing image rather than specifying it each time

### DIFF
--- a/rca/landingpages/utils.py
+++ b/rca/landingpages/utils.py
@@ -7,10 +7,10 @@ def news_teaser_formatter(item, image=None):
     editorial_type = item.editorial_types.first()
     item_as_dict["type"] = editorial_type.type if editorial_type else ""
 
-    image = get_listing_image(item)
-    if image:
-        item_as_dict["image"] = image.get_rendition("fill-392x284").url
-        item_as_dict["image_alt"] = image.alt
+    listing_image = get_listing_image(item)
+    if image and listing_image:
+        item_as_dict["image"] = listing_image.get_rendition("fill-526x402").url
+        item_as_dict["image_alt"] = listing_image.alt
     item_as_dict["formatted_date"] = item.published_at.strftime("%-d %B %Y")
     item_as_dict["title"] = item.title
     item_as_dict["link"] = item.url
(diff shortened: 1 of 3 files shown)


rca/landingpages/utils.py
```diff
@@ -1,13 +1,17 @@
...
 
     editorial_type = item.editorial_types.first()
     item_as_dict["type"] = editorial_type.type if editorial_type else ""
 
-    if image and item.hero_image:
-        item_as_dict["image"] = item.hero_image.get_rendition("fill-392x284").url
-        item_as_dict["image_alt"] = item.hero_image.alt
-    item_as_dict["formatted_date"] = item.published_at
+    listing_image = get_listing_image(item)
+    if image and listing_image:
```
richbrennan, line 11: What is `image` supposed to be here?

rca/utils/models.py
```diff
@@ -725,3 +725,21 @@ def clean(self):
...
         )
     ]
+
+
+def get_listing_image(page):
+    """Global function to get a page listing image, should use the
+    listing_image if set, if not, check for a hero image
+
+    Args:
+        page (wagtail.code.Page): The page object
```
richbrennan, line 735: Wagtail `Page` is unlikely to have a listing or hero image, maybe remove `(wagtail.code.Page)(wagtail.code.Page)`

rca/utils/models.py
```diff
@@ -725,3 +725,21 @@ def clean(self):
...
+
+    Returns:
+        rca.image.CustomImage / None: The image object
+    """
+    image = None
+    if hasattr(page, "listing_image") and page.listing_image:
+        image = page.listing_image
+    elif hasattr(page, "hero_image") and page.hero_image:
+        image = page.hero_image
+    return image
```
richbrennan, line 745: You can probably reduce this code down to:

```
image = getattr(page, "listing_image")
if not image:
    image = getattr(page, "hero_image") 
return image
```

richbrennan (changes requested):
A few comments, thanks Kevin

rca/landingpages/utils.py
kevinhowbrook: `image` is passed in optionally, if you want an image for the teaser, eg here https://github.com/torchbox/rca-wagtail-2019/blob/p3.2/sprint-2/rca/landingpages/models.py#L672 because only the one item here has an image, and there's no point sending images through that aren't going to be used

richbrennan: Ah, so `image` is a boolean flag to indicate that an image is to be returned, rather than an image itself?

In that case, for readability I'd consider renaming `image` to something like `return_image` and making it default to `False` rather then `None`

richbrennan (changes requested):

PR #737, opened 2021-08-03
883 -Event Listing Page fields and listing logic

kevinhowbrook:
Ticket [883](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/883)

Implements fields and listing logic for the event index page.

This PR also adds a new method to the event detail page for listing out the dates in a very specific way without interrupting the existing template declarations for other page dates (now or in the future)


rca/events/models.py
```diff
@@ -26,14 +27,132 @@
...
-from .forms import EventPageAdminForm
+from .forms import EventIndexPageAdminForm, EventPageAdminForm
 
 
-class EventIndexPage(BasePage):
+class EventIndexPageRelatedEditorialPage(Orderable):
+    page = models.ForeignKey(
+        "events.EventDetailPage",
+        null=True,
+        blank=False,
```
richbrennan, line 37: Shouldn't this be required and not nullable?

rca/events/forms.py
```diff
@@ -11,3 +11,7 @@ def __init__(self, *args, **kwargs):
         self.fields["contact_model_image"].label = "Image"
         self.fields["contact_model_text"].label = "Text"
         self.fields["contact_model_link_text"].label = "Link text"
+
+
+class EventIndexPageAdminForm(EventPageAdminForm):
+    pass
```
richbrennan, line 17: Will this be added to later?

richbrennan (commented):
A couple of comments, thanks Kevin!

rca/events/forms.py
kevinhowbrook: Nope - it's pretty much the same as `EventPageAdminForm` for now, so rather than declare the same thing I've inherited it

rca/events/models.py
kevinhowbrook: Yep - thanks

rca/events/forms.py
richbrennan: In that case could it be renamed something like `EventAdminForm` and used on both pages, kinda feels like a misuse of inheritance again.

PR #739, opened 2021-08-04
Feature/787 event listing filters

kevinhowbrook:
**Contains #737**

Ticket [787](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/787)

This PR Implements the filters for the listing page in commit 1dcb0ae939837457f3c5d1127257419b3df7e894

I didn't spot this in review but we already have related schools and
centres configured to relate to the Page, so we can expose those panels,
this means we don't have to rewrite the filter to work with different
relationship names


richbrennan (commented):
Just one about multiple migrations, thanks Kevin

PR #757, opened 2021-08-16
Moving main-content to h1 tag on all templates

nicklee:
Moving main content id onto the h1, so screen readers don't get taken to the breadcrumb by default (when selecting skip to content)


rca/project_styleguide/templates/patterns/pages/editorial/editorial_detail.html
```diff
@@ -25,9 +25,7 @@
                     <nav class="title-area__breadcrumb" aria-label="Breadcrumb">
                         {% include "patterns/molecules/breadcrumb/breadcrumb.html" %}
                     </nav>
-                    {% if page.title %}
-                        <h1 class="title-area__heading heading heading--display-two">{{ page.title }}</h1>
-                    {% endif %}
+                    <h1 class="title-area__heading heading heading--display-two" id="main-content">{{ page.title }}</h1>
```
chris-lawton, line 28: i'm a bit out of the loop with the structure of the site now but I presume moving `id="main-content"` doesn't mean that there's a chance that there will be more than 1 of them on the page?

chris-lawton: i'm guessing not as they're going on the `<h1>`'s...

chris-lawton (approved):

chris-lawton:
Also wondering if this is something we should adopt in kit if it's best practice

rca/project_styleguide/templates/patterns/pages/editorial/editorial_detail.html
nicklee: yea, only pages that have multiple h1's are ones with modals on, which wouldn't need the anchor on anyhow, so think we're good. Does feel annoying that we need to move this id into the templates instead of using base, but needs must…

nicklee:
> Also wondering if this is something we should adopt in kit if it's best practice

Tricky isn't it, I guess really you'd want the breadcrumb to sit outside of main content so you could still use the WK normal setup

PR #761, opened 2021-08-16
Feature/939 skip nav fix rebase

nicklee:


nicklee:
Cherry picked from https://github.com/torchbox/rca-wagtail-2019/pull/757

PR #784, opened 2021-08-24
888 Adds an add to calendar link to download an ics file

kevinhowbrook:
Ticket [888](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/888)


rca/events/models.py
```diff
@@ -618,6 +621,75 @@ def get_booking_bar(self):
             ),
         }
 
+    def add_slashes(self, string):
```
tombola, line 624: Suggestion: this utility function could do with a more descriptive name, eg `ical_escape_characters` and a comment.

```
"""
Escapes special characters in long form text fields for iCal records.
"""
```

rca/events/models.py
```diff
@@ -618,6 +621,75 @@ def get_booking_bar(self):
...
+        string.replace('"', '\\"')
+        string.replace("\\", "\\\\")
+        string.replace(",", "\\,")
+        string.replace(":", "\\:")
+        string.replace(";", "\\;")
+        string.replace("\n", "\\n")
+        return string
+
+    def make_ics(self):
```
tombola, line 633: nitpick: `get_ical_record`?

tombola (commented):
LGTM in general.

iCal functionality _could_ be separated out of event and into a couple of functions in its own file so it is available for re-use, but this might be pre-optimisation for no good reason :)

PR #791, opened 2021-08-26
Feature/893 legacy new and events

kevinhowbrook:
Ticket [893](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/893) 

image


rca/utils/models.py
```diff
@@ -396,6 +398,148 @@ class LegacySiteTaggedPage(ItemBase):
...
+            if editorial_published_date:
+                editorial_published_date = editorial_published_date.strftime(
+                    "%-d %B %Y"
+                )
+
+            PAGE_DESCRIPTION_MAPPING = {
+                "EditorialPage": editorial_published_date,
+                "EventDetailPage": getattr(page, "event_date_short", None),
+            }
+            description = PAGE_DESCRIPTION_MAPPING.get(page.__class__.__name__, "")
```
tombola, line 461: It looks like all of this mapping (and the complication with editorial type above) are the result of this function taking two different page types, and using different fields on each to populate `meta` and `description` fields.

Would it make sense to instead add a `listing_meta` and `listing_description` property to each of these models, and return them in the same way as `page.url` etc?

If not, alternatively you could just do an if, or a single map (with the values being methods) and call eg `format_editorial_page` to return a dictionary, and use that to `.update` the appended item? That way you wouldn't have to consider both page types within this generic function.

rca/utils/models.py
```diff
@@ -437,7 +581,15 @@ def legacy_news_and_events(self):
         ``page.legacy_news_and_events`` can be referenced in the
         template multiple times without triggering another
         cache lookup.
+
+        If there is no ``page.legacy_news_and_event_tags`` value, it returns
+        Editorial and Event pages with relationships to this page via
+        utils.NewsAndEventsMixin.
         """
+
+        if not self.legacy_news_and_event_tags.exists():
```
tombola, line 590: praise: quick performative check

rca/utils/models.py
```diff
@@ -437,7 +581,15 @@ def legacy_news_and_events(self):
...
+            return NewsAndEventsMixin(page=self).get_data()
```
tombola, line 591: praise: neat, I like the forethought here, using the new functionality from the start, but keeping it separate, rather than mashing it into the old mixin and having to unpick it later

tombola (commented):
Essentially LGTM, I have left some comments, mainly some suggested refactors for formatting items, intent being readability/maintainability

rca/utils/models.py
```diff
@@ -396,6 +398,148 @@ class LegacySiteTaggedPage(ItemBase):
...
+    """
+
+    def __init__(self, page, *args, **kwargs):
+
+        from rca.landingpages.models import LandingPage
+        from rca.events.models import EventDetailPage
+        from rca.editorial.models import EditorialPage
+
+        self.event_page = EventDetailPage
+        self.editorial_page = EditorialPage
```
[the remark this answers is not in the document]
kevinhowbrook: Thanks, I'm loading them here to avoid the circular import, I'll rename them though; that's a good nitpick!

rca/utils/models.py
```diff
@@ -396,6 +398,148 @@ class LegacySiteTaggedPage(ItemBase):
...
+            related_landing_pages__page=self.page
+        ).live()[: self.editorial_items]
+        return list(chain(news, event))
+
+    def get_research_page_editorial_and_events(self):
+        event = (
+            self.event_page.objects.live()
+            .filter(related_research_centre_pages__page=self.page)
+            .filter(start_date__gte=timezone.now().date())
+            .order_by("-start_date")[:1]
```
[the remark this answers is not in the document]
kevinhowbrook: Thanks, good spot!

rca/utils/models.py
kevinhowbrook: Yea we've sorted of written ourselves into a corner with this.
It needs to be one generic method as the QS we pass to be formatted is chained and will either be:
[editorial,editorial,event]
or
[editorial,editorial,editorial]

I don't see any blocking issues here rather than readability / preference so, given the time left, I'm going to leave this as is.

tombola: absolutely - not sure I follow the chaining aspect, but you have clearly given it some thought so happy with that

PR #793, opened 2021-09-01
Work in student > supervisor relationship

kevinhowbrook:
Addresses feedback on [891](https://projects.torchbox.com/projects/rca-website-rebuild/tickets/891#update-68099199)

### DIFF
--- a/rca/people/models.py
+++ b/rca/people/models.py
@@ -322,9 +322,27 @@ class StaffPage(BasePage):
                 # the value next time it runs
                 pass
 
+    def format_student_page(self, page):
+        student_page = page
+        image = getattr(student_page, "profile_image", None)
+        if image:
+            image = image.get_rendition("fill-60x60").url
+        return {
+            "first_name": student_page.first_name,
+            "surname": student_page.last_name,
+            "status": student_page.degree_status,
+            "link": student_page.url,
+            "image_url": image,
+        }
+
     def get_related_students(self):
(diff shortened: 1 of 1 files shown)


rca/people/models.py
```diff
@@ -337,21 +355,20 @@ def get_related_students(self):
...
-                item = {
-                    "first_name": student_page.first_name,
-                    "surname": student_page.last_name,
-                    "status": student_page.degree_status,
-                    "link": student_page.url,
-                    "image_url": image,
-                }
+                item = self.format_student_page(student_page)
             else:
                 item = {
```
tombola, line 373: nitpick: If the above is a student page, then this is this a student _user_?

could it also have a `format_*` function, if only so both are self documenting?

rca/people/models.py
```diff
@@ -337,21 +355,20 @@ def get_related_students(self):
...
-                if image:
-                    image = image.get_rendition("fill-60x60").url
-                item = {
-                    "first_name": student_page.first_name,
-                    "surname": student_page.last_name,
-                    "status": student_page.degree_status,
-                    "link": student_page.url,
-                    "image_url": image,
-                }
+                item = self.format_student_page(student_page)
```
tombola, line 371: nitpick: could `continue` rather than the `else`, but that is just aesthetic :)

tombola (approved):
LGTM, just random opinions

PR #801, opened 2021-10-15
Don't bother trying to make a link if there is no target

kevinhowbrook:
If a page is deleted, and the navigation item is updated, there is
potential to throw NoneType errors

### DIFF
--- a/rca/navigation/templatetags/navigation_tags.py
+++ b/rca/navigation/templatetags/navigation_tags.py
@@ -1,7 +1,10 @@
+import logging
+
 from django import template
 
 from rca.navigation.templatetags import utils
 
+logger = logging.getLogger(__name__)
 register = template.Library()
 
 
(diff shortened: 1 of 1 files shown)


rca/navigation/templatetags/navigation_tags.py
```diff
@@ -45,6 +45,8 @@ def primarynav(context):
                 tertiary_links = []
                 if tertiary_pages:
                     for i in tertiary_pages:
+                        if not i["page"] and not i["url"]:
+                            continue
```
alxbridge, line 49: **suggestion (non-blocking)** Is there any worth in logging a warning here, or would that information not feed back anywhere helpful?

alxbridge:
Change looks good for preventing a recurrence of the issue :+1:

rca/navigation/templatetags/navigation_tags.py
kevinhowbrook: :+1:

PR #816, opened 2021-12-10
Update deployment instructions on readme

victoriachan:
Just to include how to trigger pipeline if deployment doesn't happen automatically.


kevinhowbrook (approved):
Thanks @victoriachan

PR #831, opened 2022-01-05
Feature/2369 enquiry form

richbrennan:
Implements: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2369

Time hasn't allowed for unit tests.

I've confirmed with Nick to return the scholarships for the form as JSON.


rca/scholarships/views.py
```diff
@@ -45,3 +56,40 @@ def scholarships_delete(request):
             "submit_url": (reverse("scholarships_delete")),
         },
     )
+
+
+def load_scholarships(request):
+    scholarships = Scholarship.objects.all()
+
+    try:
+        programme = ProgrammePage.objects.get(pk=int(request.GET.get("programme")))
```
kevinhowbrook, line 65: I think we will need this to work with slugs rather than the IDs. So links are descriptive when sent to students, EG enquire?programme=architecture

kevinhowbrook (changes requested):

rca/scholarships/models.py
```diff
@@ -56,12 +56,21 @@ class Scholarship(models.Model):
...
 
+    class Meta:
+        ordering = ("title",)
+
     def __str__(self) -> str:
-        return self.title
+        text = self.title
+        if self.location:
+            text += f" ({self.location})"
+        return text
```
kevinhowbrook, line 73: :ok_hand:

richbrennan: I had to work around the location field being added later :grimacing:

rca/scholarships/views.py
richbrennan: Done. It needs to work with both otherwise lots of form changes required...

rca/scholarships/views.py
```diff
@@ -45,3 +56,49 @@ def scholarships_delete(request):
...
+            # if programme is an int, search by programme page id
+            programme = int(programme)
+            programme_page = ProgrammePage.objects.get(id=programme)
+        except (ProgrammePage.DoesNotExist, ValueError):
+            # if not, search by slug
+            try:
+                programme_page = ProgrammePage.objects.get(slug=programme)
+            except ProgrammePage.DoesNotExist:
+                programme_page = None
+                scholarships = scholarships.none()
```
kevinhowbrook, line 76: Good fix :ok_hand:

kevinhowbrook (approved):

PR #833, opened 2022-01-06
[2372] code format

richbrennan:
Fixes taxonomy slug length being less than taxonomy title: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2372#update-68957499


rca/utils/models.py
```diff
@@ -855,9 +855,6 @@ def save(self, *args, **kwargs):
         self.slug = slugify(self.title)
         super(SluggedTaxonomy, self).save(*args, **kwargs)
 
-    class meta:
-        abstract = True
-
```
kevinhowbrook, line 860: Am I right in saying we are making this concrete because otherwise, we would have to declare it on each class that inherits SluggedTaxonomy?

richbrennan: It is already concrete, I'm removing this to avoid future confusion.

Because `meta` is all lowercase the `abstract` part isn't being picked up by Django, which looks for `Meta`. So all models that inherit from `SluggedTaxonomy` have a concrete `SluggedTaxonomy` parent as well as concrete child. I don't think we can just rename `meta` to `Meta` because the tables are already created and linked with OneToOne relationships. I'd need to investigate more.

kevinhowbrook: :exploding_head: very good catch

kevinhowbrook (approved):

PR #834, opened 2022-01-06
[2374] add scholarships anchor link

richbrennan:
Fixes: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2374#update-68958513

### DIFF
--- a/rca/scholarships/models.py
+++ b/rca/scholarships/models.py
@@ -112,16 +112,25 @@ class ScholarshipsListingPage(ContactFieldsMixin, BasePage):
     def anchor_nav(self):
         """Build list of data to be used as in-page navigation"""
         items = []
-        blocks = []
 
-        for block in self.body:
-            blocks.append(block)
-        for block in self.lower_body:
-            blocks.append(block)
-
-        for i, block in enumerate(blocks):
+        def process_block(block):
             if block.block_type == "anchor_heading":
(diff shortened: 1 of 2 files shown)


rca/scholarships/models.py
```diff
@@ -112,16 +112,25 @@ class ScholarshipsListingPage(ContactFieldsMixin, BasePage):
...
+                items.append({"title": block.value, "link": f"#{slugify(block.value)}"})
 
         for block in self.body:
-            blocks.append(block)
+            process_block(block)
+
+        # insert link for hardcoded "Scholarships for YYYY/YY" heading
+        items.append(
+            {
+                "title": "Scholarships for 2022/23 entry",
```
kevinhowbrook, line 126: This is the right change to make for now. However some 2023, this text will need updating, and a dev will have to do it.
We are creating this area in https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2376, it will be a title field from the page. I think the best course of action here is to merge this change to pass the AC, then (time willing) amend this line to use the title field that will be brought in.

kevinhowbrook (approved):

rca/scholarships/models.py
richbrennan: Gah, I'd forgotten it was going to be a dynamic field anyway...

PR #838, opened 2022-01-10
[2376] add scholarship listing fields

richbrennan:
Implements: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2376


rca/scholarships/tests/test_models.py
```diff
@@ -31,16 +31,14 @@ def test_factories(self):
 
 
 class TestScholarshipsListingPageRules(WagtailPageTests):
-    def setUp(self):
-        self.home_page = HomePage.objects.first()
```
richbrennan, line 35: This query runs for a test that doesn't need it

kevinhowbrook: yea it was used in test_singlet but I see that's be re-jigged :+1:

kevinhowbrook (approved):

PR #840, opened 2022-01-11
JS to output scholarships based on json

nicklee:
Javascript to output the JSON supplied by BE as a list of scholarships on the form page:

https://rca-development.herokuapp.com/scholarship-enquiry/

### DIFF
--- a/rca/static_src/javascript/components/scholarship-list.js
+++ b/rca/static_src/javascript/components/scholarship-list.js
@@ -1,0 +1,41 @@
+class ScholarshipList {
+    static selector() {
+        return '[data-scholarships-url]';
+    }
+
+    constructor() {
+        this.form = document.getElementById('scholarship-form');
+        this.dataURL = this.form.getAttribute('data-scholarships-url');
+        this.programmeChooser = document.getElementById('id_programme');
+
+        this.getData();
+        this.bindEvents();
+    }
+
+    getData() {
+        const xmlhttp = new XMLHttpRequest();
+        xmlhttp.onreadystatechange = function getScholarships() {
+            if (this.readyState === 4 && this.status === 200) {
+                const dataLocation = document.getElementById('id_scholarships');
+                const scholarships = JSON.parse(this.responseText);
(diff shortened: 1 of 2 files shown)


rca/static_src/javascript/components/scholarship-list.js
```diff
@@ -0,0 +1,42 @@
...
+        this.getData();
+        this.bindEvents();
+    }
+
+    getData() {
+        // Ideally this would be refactored to use fetch
+        const xmlhttp = new XMLHttpRequest();
+        xmlhttp.onreadystatechange = function getScholarships() {
+            if (this.readyState === 4 && this.status === 200) {
+                const dataLocation = document.getElementById('id_scholarships');
```
SimonDEvans, line 20: Could keep this with the other vars defined in the constructor.

rca/static_src/javascript/components/scholarship-list.js
```diff
@@ -0,0 +1,42 @@
...
+                const scholarships = JSON.parse(this.responseText);
+                const scholarshipsData = scholarships
+                    .map((item, index) => {
```
SimonDEvans, line 23: Massive nitpick, but backticks afford the ability to format the HTML so it's more readable. E.g something like this (untested):

```
.map((item, index) => {
    return `
        <li>
            <label for="id_scholarships_${index}">
                <input id="id_scholarships_${index}" type="checkbox" name="scholarships" value="${item.id}">
                ${item.title}
            </label>
        </li>`;
})
```

SimonDEvans (approved):
Would be nice to use `fetch()` as per your comment, all good otherwise 😄

rca/static_src/javascript/components/scholarship-list.js
nicklee: This is loads better, updated

rca/static_src/javascript/components/scholarship-list.js
nicklee: Agreed, but it doesn't seem to work, confused me…

PR #846, opened 2022-01-11
[2377] don't show scholarships on listing on initial page load

richbrennan:
don't show scholarships on listing on initial page load

### DIFF
--- a/rca/scholarships/models.py
+++ b/rca/scholarships/models.py
@@ -172,66 +172,67 @@ class ScholarshipsListingPage(ContactFieldsMixin, BasePage):
         context = super().get_context(request, *args, **kwargs)
         context["anchor_nav"] = self.anchor_nav()
 
-        queryset = Scholarship.objects.prefetch_related(
-            "eligable_programmes", "funding_categories", "fee_statuses"
-        )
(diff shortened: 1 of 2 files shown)


rca/scholarships/models.py
```diff
@@ -172,66 +172,67 @@ def get_context(self, request, *args, **kwargs):
...
+                queryset = f.apply(queryset, request.GET)
+
+            # Format scholarships for template
+            results = [
+                {
+                    "value": {
+                        "heading": s.title,
+                        "introduction": s.summary,
+                        "eligible_programmes": ",".join(
+                            x.title for x in s.eligable_programmes.live()
```
kevinhowbrook, line 216: I _think_ `x.title` here should be the __str__ method?
I might be wrong so I'll leave it to your judgement, but I think you/we updated the __str__ to output the programme level too? Just worth checking what that renders out like.

kevinhowbrook (approved):

rca/scholarships/models.py
richbrennan: That was you, two years to add the degree level, but I'll take that :+1:

PR #854, opened 2022-01-26
Feature/2361 separate event times

richbrennan:
Closes: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2361


kevinhowbrook:
> bonus points for spotting why a test would fail on CI but not locally

The test that's failing looks like it's dependent on a list structure, it's targeting [0], then [1] etc. I've seen this before with circle, where objects are created in different orders.

rca/events/models.py
```diff
@@ -637,7 +658,7 @@ def event_date_short(self):
 
     @property
     def past(self):
-        return self.end_date.date() < datetime.date.today()
+        return self.end_date < datetime.date.today()
```
kevinhowbrook, line 661: Would this throw a TypeError if `self.end_date` is None?

richbrennan:
> > bonus points for spotting why a test would fail on CI but not locally
> 
> The test that's failing looks like it's dependent on a list structure, it's targeting [0], then [1] etc. I've seen this before with circle, where objects are created in different orders.

On further inspection the test data are `Orderable` but no `sort_order` was set (so presumably they're all `1` or `0`), so I'm testing a fix that adds `sort_order` to the test data...

rca/events/tests/test_models.py
```diff
@@ -173,6 +174,104 @@ def test_different_year_event(self):
...
+                        end_time=end_time,
+                        path="0",
+                        depth=0,
+                        event_type=EventTypeFactory(),
+                    ).full_clean()
+
+                the_exception = cm.exception
+                self.assertEqual(1, len(the_exception.messages))
+                self.assertEqual(message, the_exception.messages[0])
+
```
kevinhowbrook, line 274: **Praise**

Nice tests carried along with new conditional code :)

rca/events/models.py
richbrennan: Yes (and would have before this PR), but `end_date` is a required field

richbrennan:
All 137 tests now pass :tada:

kevinhowbrook (approved):

rca/events/models.py
kevinhowbrook: Thanks for fixing it

richbrennan:
@kevinhowbrook 

Please see updated `0015_migrate_event_dates_to_date_and_time.py` to account for timezones

kevinhowbrook (approved):
Thanks @richbrennan I tested running this locally against the prod db and the times look consistent to me. :pray:

PR #860, opened 2022-02-01
Fix/1171 programme finder ordering

richbrennan:
Closes: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/1171

Fixes programme ordering on programme finder and school page

### DIFF
--- a/rca/wagtailapi/filters.py
+++ b/rca/wagtailapi/filters.py
@@ -16,9 +16,9 @@ class DegreeLevelFilter(filters.BaseFilterBackend):
         pks = request.GET.getlist("project", [])
 
         if pks:
-            queryset = queryset.filter(degree_level__in=pks).order_by("title").live()
+            queryset = queryset.filter(degree_level__in=pks)
 
-        return queryset
+        return queryset.order_by("title").live()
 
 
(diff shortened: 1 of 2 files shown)


rca/wagtailapi/filters.py
```diff
@@ -16,9 +16,9 @@ def filter_queryset(self, request, queryset, view):
         pks = request.GET.getlist("project", [])
 
         if pks:
-            queryset = queryset.filter(degree_level__in=pks).order_by("title").live()
+            queryset = queryset.filter(degree_level__in=pks)
 
-        return queryset
+        return queryset.order_by("title").live()
```
victoriachan, line 21: Good work finding this bug! Not a blocker, but a suggestion: I can see the same pattern on `SubjectsFilter` and `RelatedSchoolsFilter` below, and wondering if they ought to be changed as well.

victoriachan (approved):
LGTM

rca/wagtailapi/filters.py
richbrennan: I'm not aware that the other filters are affected by the same bug so I'm going to leave it for now, thanks for spotting!

nicklee (approved):
LGTM!

PR #865, opened 2022-02-23
Support-370: EventDetailPage add to calendar link fix

BrianXu20:
Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2435

Root Issue: Trying to get time from the Date object. 

EventDetailPage has `start_time` and `end_time` fields, we need to get the time information from those fields.

### DIFF
--- a/rca/events/models.py
+++ b/rca/events/models.py
@@ -763,8 +763,8 @@ class EventDetailPage(ContactFieldsMixin, BasePage):
             # Get date
             date = self.start_date + datetime.timedelta(days=day)
             # Get times
-            start_datetime = datetime.datetime.combine(date, self.start_date.time())
-            end_datetime = datetime.datetime.combine(date, self.end_date.time())
+            start_datetime = datetime.datetime.combine(date, self.start_time)
+            end_datetime = datetime.datetime.combine(date, self.end_time)
 
             # Get a location
             location = getattr(self, "location", None)
--- a/rca/events/models.py
+++ b/rca/events/models.py
@@ -739,13 +739,14 @@ class EventDetailPage(ContactFieldsMixin, BasePage):
         """
         Escapes special characters in long form text fields for ics records.
         """
-        string.replace('"', '\\"')
-        string.replace("\\", "\\\\")
-        string.replace(",", "\\,")
-        string.replace(":", "\\:")
-        string.replace(";", "\\;")
-        string.replace("\n", "\\n")
-        return string
+        if string:
+            string.replace('"', '\\"')
+            string.replace("\\", "\\\\")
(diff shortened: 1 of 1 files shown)


rca/events/models.py
```diff
@@ -763,8 +764,8 @@ def get_ics_record(self):
             # Get date
             date = self.start_date + datetime.timedelta(days=day)
             # Get times
-            start_datetime = datetime.datetime.combine(date, self.start_date.time())
-            end_datetime = datetime.datetime.combine(date, self.end_date.time())
+            start_datetime = datetime.datetime.combine(date, self.start_time)
```
victoriachan, line 767: This throws a `TypeError` (combine() argument 2 must be datetime.time, not None) when the `start_time` field is empty (it is not a mandatory field).

rca/events/models.py
```diff
@@ -763,8 +764,8 @@ def get_ics_record(self):
...
+            end_datetime = datetime.datetime.combine(date, self.end_time)
```
victoriachan, line 768: Same problem as `start_time`.

victoriachan (changes requested):
Looks like we need to provision for what happens when `start_time` and `end_time` is empty and when only one of those are present. Probably display as all day event unless both `start_time` and `end_time` are present? As it won't make sense to have a calendar event with just one of them.

rca/events/models.py
BrianXu20: Resolved in https://github.com/torchbox/rca-wagtail-2019/pull/865/commits/26b2ee410ef2d5ef2d147cea795d4f44bdbc733d

rca/events/models.py
BrianXu20: Resolved in https://github.com/torchbox/rca-wagtail-2019/commit/26b2ee410ef2d5ef2d147cea795d4f44bdbc733d

victoriachan (changes requested):
Thanks for the change. Sorry, still one more issue. With this change when I have a one day event with no strat time or end time, it adds the event to the calendar as starting and ending at 00:00. Ideally when the event has no start or end time, it should be added as 'All Day' event on the calendar. Looks like the datetime format should be `strftime("%Y%m%d")` when there is no time specified.

BrianXu20:
Thanks, @victoriachan,
I just followed what the comment said here 
https://stackoverflow.com/a/30249034
I will try to make it that way to see if it make any difference

BrianXu20:
Screen Shot 2022-03-08 at 7 31 08 PM
Now it looks like this

victoriachan (approved):
Looks good!

PR #883, opened 2022-07-04
Modify result listings QS

kevinhowbrook:
Ticket [2453](https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2453)

- On load we only want to see events from now, to furthers in the
future.
- When filtering to future we want events ordered from now to furthest
in future.


kevinhowbrook:
@nicklee could have a check of if we can proceed with this? Ignore the issues from black atm - I'm working that out

kevinhowbrook:
Needs rebase

PR #894, opened 2022-07-06
Stop caching 500s

RealOrangeOne:
When there's an issue with the upstream server, it will return a 500. Currently, these 500s may be stored in the cache, meaning the site appears down for longer than it actually is.

This PR brings the worker more in-line with our [common caching worker](https://github.com/torchbox/cloudflare-recipes/blob/master/common-caching.js), and prevents it from caching these responses (rfc2616 defines which responses should be cached).

This also puts the worker in the repo for posterity and better versioning.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,3 +2,4 @@ node_modules
 coverage
 static_compiled
 venv
+cloudflare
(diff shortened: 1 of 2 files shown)


RealOrangeOne:
RCA is now running the common caching worker directly, without the additional proxying - so this can be closed.

PR #901, opened 2022-11-15
Fix title and search field

nickmoreton:


kevinhowbrook:
Thanks Nick

### Testing notes

**Admin**
- Can see and edit titles
- Can preview

**Editor**
- Can see and edit titles
- Can preview

**Sutdent**
- Can not see title and fields I'm not supposed to 
- Can not preview (I think that's ok for this fix though)

rca/people/admin_forms.py
```diff
@@ -6,23 +7,45 @@ class StudentPageAdminForm(WagtailAdminPageForm):
...
-            for f in readonly_fields:
+            for f in disabled_fields:
                 # Wrapped in a try except as when a student uses this form, these fields are hidden
                 # See rca.people.models.StudentPage.basic_content_panels and rca.people.utils.PerUserPageMixin
                 try:
                     self.fields[f].widget.attrs["disabled"] = True
                 except KeyError:
                     pass
 
+        if self.instance.title and self.instance.slug and not user.is_student():
```
kevinhowbrook, line 30: This will probably do for now, but I think `user` can be None (tested when previewing a student page as a student) meaning this will `user.is_student` check will throw an AttributeError

Might be worth adding in a condition to see if we can actually check the `user` too

kevinhowbrook (approved):

rca/people/admin_forms.py
```diff
@@ -6,22 +7,44 @@ class StudentPageAdminForm(WagtailAdminPageForm):
...
-                # Wrapped in a try except as when a student uses this form, these fields are hidden
-                # See rca.people.models.StudentPage.basic_content_panels and rca.people.utils.PerUserPageMixin
-                try:
-                    self.fields[f].widget.attrs["disabled"] = True
-                except KeyError:
-                    pass
+
+        user = kwargs.get("for_user")
+
+        if user:
```
nickmoreton, line 13: Checking for user here seems to fix the `preview` a draft function for an admin account but not for a student account.

PR #902, opened 2022-11-15
Switch to check if student so editors and mods can edit

kevinhowbrook:


nickmoreton:
This seems to work for me.

- As a editor or superuser I can see the search form
- As a student I can't see the search form.

nickmoreton (approved):
Look good

PR #903, opened 2022-11-15
Add tests for admin UI

kevinhowbrook:
Ensure the search bar is still visible for editor roles, not just `admin`


nickmoreton (commented):
Test are running ok for me.

**Notes**: at the moment I can see the need to add a similar test for:

- Title field (class addition)
- Slug field (class addition)

PR #906, opened 2022-12-05
#2467 - BE: Enquire to study report - country and city of residence fields

patrickcuagan:
Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2467

This PR adds the city field to the exports. It also updates to show the full country name for shown for the country of residence.

Screenshot:

Screen Shot 2022-12-05 at 2 54 22 PM

PR #907, opened 2022-12-05
#2448 Adds date filter to data export for granular reports

kevinhowbrook:
This PR adds a [date range](https://github.com/wunderweiss/wagtail-rangefilter) filter to the enquire to study export. 

This is needed as heroku has a 30s response limit which is timing out when larger data sets are exported. With this filter, the administrator can create exports within a specific timeframe meaning the data will be smaller and will be within the 30s threshold.

Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2448

 <details>
  <summary>Screenshots</summary>

image

</details>

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -283,6 +283,14 @@ argon2 = ["argon2-cffi (>=19.1.0)"]
 bcrypt = ["bcrypt"]
 
 [[package]]
+name = "django-admin-rangefilter"
+version = "0.9.0"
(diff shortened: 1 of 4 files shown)


patrickcuagan (approved):
LGTM! TIL about `wagtail-rangefilter`.

The optimization we did in https://github.com/torchbox/rca-wagtail-2019/pull/906 might also help with the timeout.

PR #924, opened 2023-01-03
Install poetry using pip

nickmoreton:
The CI command that uses this url:

`curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py ...`

is no longer available and returns 404.

This updates the installation of poetry to use pip.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,8 +22,8 @@ jobs:
       - run:
           name: install python dependencies
           command: |
-            curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | POETRY_VERSION=1.1.8 python
-            . $HOME/.poetry/env
+            pip install --upgrade pip
+            pip install poetry==1.1.15
             python -m venv venv
             . ./venv/bin/activate
             poetry install


.circleci/config.yml
```diff
@@ -22,8 +22,8 @@ jobs:
       - run:
           name: install python dependencies
           command: |
-            curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | POETRY_VERSION=1.1.8 python
-            . $HOME/.poetry/env
+            pip install --upgrade pip
```
victoriachan, line 25: Have you tried replacing get-poetry.py with https://github.com/python-poetry/install.python-poetry.org? e.g. `curl -sSL https://install.python-poetry.org | POETRY_VERSION=1.1.15 python3 -`? I'm not sure if this is better than using pip install but it looks it looks more in line with the way it was done before.

.circleci/config.yml
```diff
@@ -22,8 +22,7 @@ jobs:
       - run:
           name: install python dependencies
           command: |
-            curl -sSL https://raw.githubusercontent.com/python-poetry/poetry/master/get-poetry.py | POETRY_VERSION=1.1.8 python
-            . $HOME/.poetry/env
```
nickmoreton, line 26: Thanks @victoriachan your suggestion worked. I just needed to remove this line so it's not sourcing a directory that no longer exists.

victoriachan: Great!

victoriachan (approved):

kevinhowbrook (approved):

PR #930, opened 2023-03-01
#2483 - Add TAP Widget to additional page types

patrickcuagan:
Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2483

This PR adds the TAP widget field to landing pages. It looks like landing pages already had the field since they already inherit from `LandingPage` - it just wasn't added to the content panels.

### DIFF
--- a/rca/landingpages/models.py
+++ b/rca/landingpages/models.py
@@ -1386,54 +1386,60 @@ class TapLandingPage(LandingPage):
     class Meta:
         verbose_name = "Landing Page - TAP"
 
-    content_panels = BasePage.content_panels + [
-        MultiFieldPanel(
-            [FieldPanel("hero_image")],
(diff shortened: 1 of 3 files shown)


patrickcuagan:
> Hey @patrickcuagan,
> 
> the changes look fine in terms of the BE, it's not clear what changes are required for the FE? It looks these pages need some tweaks considering the editors could not add these fields and looking at the templates they do not all inherit from the generic listing index.

Thanks Will - I've double checked and did notice that some landing pages don't have the ff block of code:

```
{% block tap_widget %}
     {% if settings.utils.SitewideTapSetting.show_widgets %}{{ tap_widget_code }}{% endif %}
{% endblock %}
```

I've added this one in now.

PR #933, opened 2023-03-03
#2490 - SPIKE: Research students unable to submit updates/content to their profiles

patrickcuagan:
Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2490

This PR fixes an issue where the school field does not exist when a student edits their own page.

Issue: We're not rendering the form fields if a student is not logged in. The formset doesn't see those fields so an error is raised.

### DIFF
--- a/rca/people/templates/admin/panels/student_page_inline_panel.html
+++ b/rca/people/templates/admin/panels/student_page_inline_panel.html
@@ -3,9 +3,11 @@
 {% comment "Always render the formset so the form can be saved" %}{% endcomment %}
 {{ self.formset.management_form }}
 
-{% comment "Hide the visibe form fields to non-superusers" %}{% endcomment %}
-{% if request.user.is_superuser %}
-
+{% comment %} 
+Hide the visible form fields to non-superusers. We only want to toggle the visibility since
+the fields must still exist in the template so it's part of the POSTed data.
+{% endcomment %}
+<div {% if not request.user.is_superuser %}style="display: none;"{% endif %}>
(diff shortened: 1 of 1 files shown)


kevinhowbrook (approved):
Thanks @patrickcuagan, if there is any time left one improvement here would be creating some more tests to visually check the presence of these fields (in case the templates are changed further)

patrickcuagan:
> Thanks @patrickcuagan, if there is any time left one improvement here would be creating some more tests to visually check the presence of these fields (in case the templates are changed further)

Thanks Kev - I've added a test now.

rca/people/tests/test_models.py
```diff
@@ -275,3 +275,60 @@ def test_create_page_validation_no_collection(self):
...
+                title="Students",
+                slug="students",
+                introduction="students",
+            )
+        )
+        self.student_index = StudentIndexPage.objects.first()
+        GroupPagePermission.objects.create(
+            group=self.student_group,
+            page=self.student_index,
+            permission_type="edit",
```
nickmoreton, line 307: Hi @patrickcuagan I needed to revert this work as it was pushed to dev and causing problem with deploying the Wagtail 5.1 upgrade to dev for testing.

My suggestion here is untested but reflects the changes required here: https://docs.wagtail.org/en/stable/releases/5.1.html#grouppagepermission-now-uses-django-s-permission-model

```suggestion
            permission=Permission.objects.get(
                content_type__app_label="wagtailcore", codename="change_page"
            )
```

It may be best to wait until the 5.1 upgrade has been deployed and the rebase your work to see the error in the test.

PR #936, opened 2023-04-26
Support/wagtail 42 upgrade

nickmoreton:
This upgrade wagtail to v4.2

Support ticket: [1166257309](https://torchbox.monday.com/boards/1130687432/pulses/1166257309)

Some of the changes here are not specifically for this upgrade and look like they are items form either version upgrades that were missed.

- Imports and use_json_field + some migrations
- wagtail_redirect_importer (removed as it's now part of wagtail since 2.10)

Additionally: 
- Check Student page inline javascript still works as expected. https://github.com/torchbox/rca-wagtail-2019/pull/936/commits/8bcefd403770ae8ed51eeddefdb9f6d93034d681


alxbridge (approved):
@nickmoreton These changes all look good, and are complete as far as I can see. I've also done some testing of the site functionality, though I've not checked all the critical paths from the upgrading doc.

PR #937, opened 2023-05-22
Upgrade poetry to 1.5.0

nickmoreton:
This upgrades poetry to v 1.5.0

Support ticket: [1198296347](https://torchbox.monday.com/boards/1143413211/pulses/1198296347)

PR: #936 is currently blocked by this.


alxbridge:
I've just pushed a commit to fix the CircleCI config, so that Poetry can be upgraded past v1.2: the issue is that CircleCI doesn't provide full terminal emulation, and newer versions of Poetry require terminal features that go beyond what it does provide.

The good news is that these features can be disabled by adding the `--no-ansi` flag to any `poetry` commands.

See https://github.com/python-poetry/poetry/issues/7184 for full details

PR #938, opened 2023-07-03
Update Django to fix CVE-2023-36053

RealOrangeOne:
I'm not entirely sure why `tablib` has updated. There's no reason for it, but it reliably happens. It's possibly an unpinned version somewhere which gets updated even when I run `poetry lock --no-update`.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -250,7 +250,7 @@ python-versions = "*"
 
 [[package]]
 name = "django"
-version = "3.2.16"
+version = "3.2.20"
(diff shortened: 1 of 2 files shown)


katdom13 (approved):
Encountered a similar issue. It's apparently a problem with the dependency solver that older versions of Poetry use (cc @alxbridge)
The approach I took was to accept the bump and do some testing.

nickmoreton:
I'm closing this because since was created Django is now at 4.2 so this isn't required.

PR #940, opened 2023-08-14
#2505 | Fix: Conditional rendering of images on Requirements and Apply tab optional

engineervix:
[Ticket](https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2505)

When a _curriculum image_ or _apply image_ is not specified on a Programme page, we expect that the image should not be rendered at all. However, what happens is that the image block is displayed with a null URL. This PR fixes this by checking that the images exist before proceeding to render them.

<details>
  <summary>Here is an illustration</summary>

https://github.com/torchbox/rca-wagtail-2019/assets/7713776/feaccdcb-ffb0-4986-b0f6-55aa95af9a54

</details>

### DIFF
--- a/rca/project_styleguide/templates/patterns/molecules/image-caption-link/image-caption-link.html
+++ b/rca/project_styleguide/templates/patterns/molecules/image-caption-link/image-caption-link.html
@@ -8,11 +8,12 @@
     {% else %}
         <div class="image-caption-link__container">
     {% endif %}
+        {% if image %}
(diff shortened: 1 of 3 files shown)


kevinhowbrook (approved):
Nice, thanks @engineervix - appreciate the checks in the parent templates too.

Just a reminder that we don't use staging on RCA, so this should go to dev first for testing and apply the label 'on-dev' to the PR.

Thanks!

PR #942, opened 2023-08-15
#2503 BE | Programme page > Create programme details duration options as a Taxonomy

engineervix:
[Ticket](https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2503)

This PR makes changes to the way the programme details duration (study mode) options are defined.

- A new Programme study mode Taxonomy is created, and is used as the basis for defining the study mode for a `ProgrammePage`. The old field `` is removed from the `ProgrammePage` model.
- A data migration ensures that the existing information is preserved.

<details>
  <summary>Here's a quick demo</summary>

https://github.com/torchbox/rca-wagtail-2019/assets/7713776/c93de993-9987-4ea3-9e0a-f77155c39b1e

</details>


rca/programmes/models.py
```diff
@@ -207,6 +208,37 @@ def __str__(self):
...
 
+class ProgrammeStudyMode(models.Model):
+    """
+    For instance, full-time, part-time, etc.
+    """
+
+    title = models.CharField(max_length=128)
+
+    def __str__(self):
+        return self.title
```
kevinhowbrook, line 219: When we come to do the filtering, this might need a slug - I'd suggest looking at the current taxonomies used to filter programmes and matching those in terms of what the model is like

kevinhowbrook: Also I would suggest we limit the creation of these instances to 2 - seen as though we will be using it as a boolean?

rca/project_styleguide/templates/patterns/molecules/key-details/key-details.html
```diff
@@ -51,9 +51,9 @@ <h4 class="body body--two key-details__sub-heading">{{ programme_page_global_fie
                         {{ page.programme_details_time }} {{ page.get_programme_details_time_suffix_display }}
                     </li>
                     {% endif %}
-                    {% if page.programme_details_duration %}
+                    {% if page.study_mode %}
                     <li class="key-details__list-item">
-                        {{ page.get_programme_details_duration_display }}
+                        {{ page.study_mode }}
```
kevinhowbrook, line 56: Does this account for a page having 2 values tagged? How would 2 render here (double check with the client but I'm pretty sure they needed to support 2 tags of full time and part time)

rca/utils/formatters.py
```diff
@@ -92,3 +92,48 @@ def format_page_teasers(obj):
...
+        # get the first part of each string (excluding the common word)
+        formatted_string = [
+            string[: -len(common_word)].strip().capitalize() for string in strings
+        ]
+
+        result = separator.join(formatted_string) + " " + common_word
+
+        return result.capitalize()
+    else:
+        return separator.join(strings).capitalize()
```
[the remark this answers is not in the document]
engineervix: Thanks @kevinhowbrook, I've moved it in 46387c61

rca/programmes/models.py
engineervix: Thanks, updated in 0f1ee3fb

kevinhowbrook (approved):

PR #948, opened 2023-08-17
#2509 BE Programme finder > Add study mode filter

engineervix:
[Ticket](https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2509)

This is to allow users to filter programmes by study mode.

Changes made:

- added a new BaseFilterBackend which checks for 'part-time'  in the API GET query params, then filters the queryset accordingly (d6edbb73)
- added a `StudyModeToggleSwitch` React component (8920b594)
- added logic to handle events on the toggle switch  (b7aa3f49)
- updated the frontend API logic to include filtering results by part-time (0f636413)

<details>
  <summary>Illustration</summary>

https://github.com/torchbox/rca-wagtail-2019/assets/7713776/d0cc5936-a196-4fb1-8a5e-af8758804537

</details>

>**Warning**
>
> - ~~The JS test suite is broken, needs fixing, @nicklee would you please take a look?~~ fixed in b539acdc
> - The toggle component state needs fixing

### DIFF
--- a/rca/static_src/javascript/main.entry.js
+++ b/rca/static_src/javascript/main.entry.js
@@ -1,37 +1,38 @@
 import '@babel/polyfill';
 import Alpine from 'alpinejs';
 
+import Accordion from './components/accordion';
+import ActualHeight from './components/actual-height';
+import AnchorNav from './components/anchor-nav';
+import BackLink from './components/back-link';
+import Carousel from './components/carousel';
+import PeekFullCarousel from './components/carousel-full-peek';
+import PeekCarousel from './components/carousel-peek';
+import CookieWarning from './components/cookie-message';
+import EmailShare from './components/email-share';
+import EventToggleSwitch from './components/event-toggle-switch';
+import FormFocus from './components/form-focus';
 import GridSizeVariables from './components/grid-size-variables';
 import HeaderDrawer from './components/header-drawer';
-import SubMenu from './components/submenu';
-import BackLink from './components/back-link';
-import MobileSubMenu from './components/mobile-sub-menu';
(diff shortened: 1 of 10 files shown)


rca/static_src/javascript/main.entry.js
engineervix, line 1: Apologies for the apparent confusion here, the only real change is the `ProgrammeToggleSwitch` import. The rest of the "changes" are just a re-organisation of the imports in alphabetical order, something that I think my code editor did when I saved the file.

rca/static_src/javascript/programmes/components/ProgrammesExplorer.js
```diff
@@ -28,6 +29,12 @@ const ProgrammesExplorer = ({ searchLabel, categories }) => {
     return (
         <>
             <SearchForm searchQuery={searchQuery} label={searchLabel} />
+            {/* Study mode toggler placeholder */}
+            <ToggleSwitch
+                ariaLabel="Programme study mode"
+                labelOne="Full-time"
+                labelTwo="Part-time"
+            />
```
engineervix, line 37: According to the design, this component shouldn't be here. I just placed it here as it was easier and quicker for me as I tested things out.

rca/static_src/javascript/programmes/programmes.api.js
engineervix, line 1: The idea here is that we always send `part-time=something` to the wagtail API endpoint, since, at any given time, we want to show programmes / short courses by their study mode (full time or part time).

Example:

When the page loads, the toggle is set to full time, and there is no query string, therefore, we send `part-time='false'` to the API endpoint.

PR #956, opened 2023-08-22
Feature/2509 fe study mode filter

nicklee:
FE and BE for Study mode filter.

Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2519

Still require some layout styling, but logic all there.

Note: We may want to remove the original js that was added as it's now all done in react.

### DIFF
--- a/rca/static_src/sass/components/_toggle-switch.scss
+++ b/rca/static_src/sass/components/_toggle-switch.scss
@@ -104,14 +104,40 @@
         .filter-bar--stuck & {
             border-color: $color--white-24;
         }
+
+        &::before {
+            content: '';
+            position: absolute;
+            top: 0;
+            left: -20px;
+            height: 1px;
+            width: 11px;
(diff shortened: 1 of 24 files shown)


rca/static_src/sass/components/_toggle-switch.scss
```diff
@@ -102,6 +106,15 @@
         }
     }
 
+    .programmes-categories & {
```
patrickcuagan, line 109: It seems like the styles for this one is specific to programmes-categories so I've implemented it like this.

nicklee:
I have a feeling we can remove "programme-toggle-switch.js" now, as the logic is done in the react component?

chris-lawton:
> I have a feeling we can remove "programme-toggle-switch.js" now, as the logic is done in the react component?

If I search the codebase for `data-study-mode-toggle` the only result I get is from that file so think it can go 👍

PR #958, opened 2023-10-06
Fix/google spam

alxbridge:
Addresses the Google spam issue as recommended on [this doc](https://docs.google.com/document/d/1sZ2q6KbiXtpIA89zQHpPC5PuTq9dl0XGr7DyzGtWmWA/edit):

- Adds a noindex meta tag to the search page
- Removes the search page disallow from robots.txt, to allow Google to pick up on the noindex change

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1,3 +1,2 @@
 User-Agent: *
-Disallow: /search/
 Allow: /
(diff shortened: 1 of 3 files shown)


public/robots.txt
```diff
@@ -1,3 +1,2 @@
 User-Agent: *
-Disallow: /search/
```
RealOrangeOne, line 2: Apparently, this `robots.txt` isn't actually being served. Either way, this is fine.

PR #960, opened 2023-10-30
Remove cooke banner

cassidydvl:
When applied this MR will remove the html, css and js for the cookie banner is favour of implementing a new cookie banner through GTM and OneTrust.

Note this task was only to remove the existing banner, the rest will be managed by analytics team.

Ticket: [2521 FE | Add GTM and One Trust cookie banner to RCA main site](https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2521)

PR #961, opened 2023-11-08
Adds CookieButtonSnippet

kevinhowbrook:
:speaker: Based off https://github.com/torchbox/rca-wagtail-2019/pull/960

This PR adds the back end support to create a cookie button snippet and add it to the body on the guide page.

To test: go to a settings > snippets > add a cookie button snippet (only one is permitted)
Then, edit a guide page and use the cookie button snippet block.

Rendering EG:

image

At the moment, I'm not sure if we need to add any javascript to this yet so I will push it to dev and see.


jams2 (approved):
The changes make sense to me @kevinhowbrook :+1:

PR #963, opened 2023-11-17
Support/wagtail 51

nickmoreton:
# Upgrade Wagtail to v5.1

Support Ticket: https://torchbox.monday.com/boards/1130687432/pulses/1264169042

This work has the latest master branch plus https://github.com/torchbox/rca-wagtail-2019/pull/939 merged in.

We are expecting: https://github.com/torchbox/rca-wagtail-2019/pull/939 to be merged and deployed before this work is merged.

Code review will be required from commit: d4dd1e5ee1023351eb5df6fe93df867fa3d90f9b

### DIFF
--- a/rca/people/utils.py
+++ b/rca/people/utils.py
@@ -117,9 +117,12 @@ class StudentPageSettingsTab(ObjectList):
                 for child in self.panel.children:
                     if child.__class__.__name__ == "PublishingPanel":
                         for field_row_panel in child.children:
-                            # Theres a FieldRowPanel inside the PublishingPanel
-                            for panel in field_row_panel.children:
-                                panel.permission = permission
+                            if field_row_panel.__class__.__name__ == "FieldRowPanel":
+                                # Theres a FieldRowPanel inside the PublishingPanel
+                                field_row_panel.permission = permission
+                                # Since Wagtail 5.1, a FieldRowPanel would have no children here
+                                # The bugfix here: https://docs.wagtail.org/en/stable/releases/5.1.1.html#bug-fixes
+                                # seems to be when it was changed.
                     # the elif's below are not required, but are here for clarity
                     elif child.__class__.__name__ == "PrivacyModalPanel":
                         pass  # Do nothing so they are still visible
--- a/poetry.lock
+++ b/poetry.lock
@@ -494,13 +494,13 @@ files = [
 
 [[package]]
 name = "django"
-version = "3.2.21"
+version = "3.2.23"
 description = "A high-level Python Web framework that encourages rapid development and clean, pragmatic design."
 optional = false
 python-versions = ">=3.6"
 files = [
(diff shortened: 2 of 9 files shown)


rca/people/utils.py
```diff
@@ -117,9 +117,12 @@ def __init__(self, **kwargs):
...
                         for field_row_panel in child.children:
-                            # Theres a FieldRowPanel inside the PublishingPanel
-                            for panel in field_row_panel.children:
-                                panel.permission = permission
+                            if field_row_panel.__class__.__name__ == "FieldRowPanel":
+                                # Theres a FieldRowPanel inside the PublishingPanel
+                                field_row_panel.permission = permission
+                                # Since Wagtail 5.1, a FieldRowPanel would have no children here
+                                # The bugfix here: https://docs.wagtail.org/en/stable/releases/5.1.1.html#bug-fixes
+                                # seems to be when it was changed.
```
nickmoreton, line 125: Since Wagtail 5.1, a FieldRowPanel would have no children here
The bugfix here: https://docs.wagtail.org/en/stable/releases/5.1.1.html#bug-fixes
seems to be when it was changed.

The custom work here, to hide the panel for permissions other than superuser,
can remain with this change.

nickmoreton:
> Hi @nickmoreton , When I pull data from staging and do a migrate command, I get the error:
> 
> ```
> django.db.migrations.exceptions.InconsistentMigrationHistory: Migration projects.0024_remove_hero_colour_options is applied before its dependency projects.0023_add_pdf_link_text on database 'default'.
> ```
> 
> Do you have any workarounds for this? Thanks!

I think you should be pulling dev data `fab pull-dev-data`

katdom13 (approved):
Hi @nickmoreton ,
LGTM! Thanks.

katdom13:
Hi @nickmoreton ,
I forgot, but it might also be a good idea to ditch the `wagtail-accessibility` package in favor of Wagtail's built-in accessibility checker in place since 4.2

https://github.com/torchbox/rca-wagtail-2019/blob/51e2b3ddff8c7ee9b747a9728df7c2e0cbf239d9/pyproject.toml#L24

While it seems to be part of pyproject.toml, I don't think it's being used anywhere else.
Thanks!

nickmoreton:
> Hi @nickmoreton , I forgot, but it might also be a good idea to ditch the `wagtail-accessibility` package in favor of Wagtail's built-in accessibility checker in place since 4.2
> 
> https://github.com/torchbox/rca-wagtail-2019/blob/51e2b3ddff8c7ee9b747a9728df7c2e0cbf239d9/pyproject.toml#L24
> 
> While it seems to be part of pyproject.toml, I don't think it's being used anywhere else. Thanks!

Hi @katdom13 I also couldn't find any usage in the site. For reference incase this pops up later the wagtail docs do mention to this package: https://docs.wagtail.org/en/stable/advanced_topics/accessibility_considerations.html#wagtail-accessibility but I can't see that the template tag is in use now. So i'm happy for it to be removed from the dependencies. 👍

PR #964, opened 2023-11-19
№ 2536 | Add optional floating CTA to guide page template

engineervix:
[ticket URL](https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2536)

This PR updates the GuidePage model to allow publishers to add a sticky CTA similar to the one used on Short Course Pages such as [this one](https://www.rca.ac.uk/study/programme-finder/designing-for-an-ageing-population/) as the **booking bar component**.

The above is achieved by creating a `StickyCTAMixin` in `rca.utils.models` , which is inherited by `rca.guides.models.GuidePage`.

A new *sticky-cta* template based on the *booking-bar* template is created, and `include`d in `templates/patterns/pages/guide/guide.html`.

Some custom validation is added to ensure that

- only a page or link is specified, and not both
- if a link is specified, a title is also specified, and vice-versa
- the description is specified if either a page or link has been specified.

> :warning: **Issue**: <br>
Some bit of FE work is required to fix the positioning of the sticky CTA (see illustration below)

<details><summary>Illustration</summary>

https://github.com/torchbox/rca-wagtail-2019/assets/7713776/d6edb45b-e47d-4e0b-a786-1a780be6981e

</details>


rca/guides/models.py
```diff
@@ -125,11 +128,46 @@ def get_related_pages(self):
...
+            )
+
+        if self.sticky_cta_page or self.sticky_cta_link_url:
+            # Either page or link URL is specified, so description is required
+            if not self.sticky_cta_description:
+                message = "You must specify a description for the sticky CTA"
+                raise ValidationError(
+                    {"sticky_cta_description": ValidationError(message)}
+                )
+
```
kevinhowbrook, line 163: **thought**

As this is a mixing, is it possible to define the validation on the StickyCTA class itself then call super().clean ?

kevinhowbrook (approved):
Thanks Victor, this is exactly what we planned :+1:

rca/guides/models.py
engineervix: I initially did this by overriding the `clean()` method on the StickyCTA class, but it never worked.

However, in 3ec923c3 I have added the validation in the `full_clean()` method and this is working

PR #965, opened 2023-11-20
#2477 - FE | Styling for Donation Form

patrickcuagan:
Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2477

This merge request updates the styling to the donation form.

Testing instructions:

1. Visit the endpoint `/pattern-library/pattern/patterns/pages/donate/donate--pattern-library-only.html/` in the pattern library.

Screenshot:

Donate


chris-lawton:
The site has some existing form styles which can be seen below or [on staging](https://rca-staging.herokuapp.com/phase-3/register-interest-fs/)

Screenshot 2023-11-20 at 10 51 06

Just wanted to check we're not meant to be using those here?

Changes look good otherwise @patrickcuagan 👍

patrickcuagan:
Yup that's right - we won't use those existing form styles for the donation widget.

PR #967, opened 2023-11-21
#2534 - FE | Staggered card module - Highlight Functionality of Links

patrickcuagan:
Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2534

This MR adds some icons to the link of the staggered card.

Testing instructions:

1. Pull data and visit the following link: `/study/application-process/`

Screenshot:

Screenshot 2023-11-21 at 9 54 49 AM

### DIFF
--- a/rca/project_styleguide/templates/patterns/molecules/card/card.html
+++ b/rca/project_styleguide/templates/patterns/molecules/card/card.html
@@ -20,7 +20,7 @@
                          <span class="card__heading-link">
                              {{ item.title }}
                              {% if link_style  == "with-arrow" %}
-                             <svg class="card__heading-link-icon" width="12" height="8">
+                             <svg class="card__heading-link-icon" width="12" height="8" aria-hidden="true">
                                  <use xlink:href="#arrow"></use>
                              </svg>
                              {% endif %}
--- a/rca/project_styleguide/templates/patterns/molecules/card/card.html
+++ b/rca/project_styleguide/templates/patterns/molecules/card/card.html
@@ -16,9 +16,18 @@
                     <div class="card__meta">{{ item.meta }}</div>
                 {% endif %}
                 <h3 class="card__heading">
-                    {% if item.link %}<span class="card__heading-link">{% endif %}
-                        {{ item.title }}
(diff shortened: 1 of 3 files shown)


rca/project_styleguide/templates/patterns/molecules/card/card.html
```diff
@@ -16,9 +16,18 @@
...
+                            {% if link_style  == "with-arrow" %}
+                            <svg class="card__heading-link-icon" width="12" height="8">
+                                <use xlink:href="#arrow"></use>
+                            </svg>
+                            {% endif %}
+                        </span>
+                    {% else %}
                         {{ item.title }}
-                    {% if item.link %}</span>{% endif %}
+                    {% endif %}
```
patrickcuagan, line 30: We could've just added just the SVG though it could've gotten a bit confusing. I've just updated this so that we just repeat `{{ item.title }}` instead.

It used to be:

```
{% if item.link %}<span class="card__heading-link">{% endif %}
    {{ item.title }}
{% if item.link %}</span>{% endif %}
```

rca/project_styleguide/templates/patterns/molecules/card/card.html
```diff
@@ -16,9 +16,18 @@
...
                 {% endif %}
                 <h3 class="card__heading">
-                    {% if item.link %}<span class="card__heading-link">{% endif %}
-                        {{ item.title }}
-                    {% if item.link %}</span>{% endif %}
+                    {% if item.link %}
+                         <span class="card__heading-link">
+                             {{ item.title }}
+                             {% if link_style  == "with-arrow" %}
+                             <svg class="card__heading-link-icon" width="12" height="8">
```
chris-lawton, line 23: I don't think this would be announced by most screen readers but it could be worth adding `aria-hidden="true"` for the ones that will (I think Narrator on Windows might).

rca/project_styleguide/templates/patterns/molecules/card/card.html
chris-lawton: Makes sense 👍

PR #979, opened 2024-01-17
Wagtail 5.2

katdom13:
# Wagtail 5.2 ([See release notes](https://docs.wagtail.org/en/stable/releases/5.2.html))

## When deployed, run the following management commands

```bash
python manage.py rebuild_references_index
python manage.py update_index
python manage.py copy_daily_hits_from_wagtailsearch
```

PR #978, opened 2024-01-16
Django 4.2

katdom13:
This PR:
- Upgrades Django from ~3.2 to ~4.2
- Upgrades other packages
- Adds new migration files after Django upgrades

PR #980, opened 2024-01-22
Integration/wagtail django upgrade

katdom13:
This integrates the following upgrades:
- [x] [Wagtail 5.2](https://github.com/torchbox/rca-wagtail-2019/pull/979)
- [x] [Django 4.2](https://github.com/torchbox/rca-wagtail-2019/pull/978)

**Important:**
This branch should always be in sync with the master branch


rca/utils/templatetags/util_tags.py
```diff
@@ -106,6 +105,9 @@ def social_media_links(context):
     "beta.rca.ac.uk",
 ]
 
+if hasattr(settings, "WAGTAILADMIN_BASE_URL"):
+    default_domains.append(settings.WAGTAILADMIN_BASE_URL)
+
```
nickmoreton, line 110: At build time WAGTAILADMIN_BASE_URL isn't set but at run time it should be. There's an env var of `PRIMARY_HOST` available.

PR #996, opened 2024-04-12
#2563 - Site-wide > Fonts > Implementation

patrickcuagan:
Ticket: https://projects.torchbox.com/projects/rca-django-cms-project/tickets/2563

This PR updates the font family used throughout the build to "Source Sans 3". 

I've also had to update the font weights as it was controlled by the font-family before. Now, we can just update the font weight.

### DIFF
--- a/rca/static_src/sass/abstracts/_variables.scss
+++ b/rca/static_src/sass/abstracts/_variables.scss
@@ -40,7 +40,7 @@ $color--meta-lightbg: $color--dark-grey;
 
 $color--focus: $color--blue;
 
-// --------------------------------- Fonts --------------------------------------
+// --------------------------------- Font faces -------------------------------------
 
 /* source-sans-3-latin-400-normal */
 @font-face {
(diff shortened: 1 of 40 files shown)


rca/static_src/sass/abstracts/_variables.scss
```diff
@@ -40,18 +40,75 @@ $color--meta-lightbg: $color--dark-grey;
 
 $color--focus: $color--blue;
 
+// --------------------------------- Font faces -------------------------------------
+
+/* source-sans-3-latin-400-normal */
+@font-face {
+    font-family: 'Source Sans 3';
+    font-style: normal;
+    font-display: swap;
```
siimonevans, line 49: I wonder if `block` would be more appropriate to prevent FOUT? 

I accept the argument that `swap` displays text on the page more quickly on first load, but it can be jarring as the type shifts once the desired font is loaded in.

rca/static_src/sass/abstracts/_variables.scss
```diff
@@ -40,18 +40,75 @@ $color--meta-lightbg: $color--dark-grey;
...
+    font-weight: 400;
+    src: url(https://cdn.jsdelivr.net/fontsource/fonts/source-sans-3@latest/latin-400-normal.woff2)
```
siimonevans, line 51: Curious to know why we didn't go for the usual Google font CDN? Don't think it matters much though.

rca/static_src/sass/abstracts/_variables.scss
```diff
@@ -40,18 +40,75 @@ $color--meta-lightbg: $color--dark-grey;
...
+            format('woff2'),
+        url(https://cdn.jsdelivr.net/fontsource/fonts/source-sans-3@latest/latin-400-normal.woff)
+            format('woff');
+    unicode-range: U+0000-00FF, U+0131, U+0152-0153, U+02BB-02BC, U+02C6, U+02DA,
```
siimonevans, line 55: Interesting! I note that in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range):
> If the page doesn't use any character in this range, the font is not downloaded; if it uses at least one, the whole font is downloaded.

Does this mean that the font will basically always be downloaded in its entirety? Perhaps I'm misunderstanding, but I'm struggling to understand the performance benefit.

rca/static_src/sass/abstracts/_variables.scss
```diff
@@ -40,18 +40,75 @@ $color--meta-lightbg: $color--dark-grey;
...
+
+/* source-sans-3-latin-400-normal */
+@font-face {
+    font-family: 'Source Sans 3';
+    font-style: normal;
+    font-display: swap;
+    font-weight: 400;
+    src: url(https://cdn.jsdelivr.net/fontsource/fonts/source-sans-3@latest/latin-400-normal.woff2)
+            format('woff2'),
+        url(https://cdn.jsdelivr.net/fontsource/fonts/source-sans-3@latest/latin-400-normal.woff)
```
siimonevans, line 53: Do we need IE11 support? If not, we can get away with just the `woff2` I think.

PR #1003, opened 2024-05-14
#3879 - Listing pages: duplicates appearing in results

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1472503879

This PR fixes an issue where duplicate results are shown when doing multiple filters in the editorial listing page. It's a simple fix!

### DIFF
--- a/rca/editorial/models.py
+++ b/rca/editorial/models.py
@@ -552,6 +552,8 @@ class EditorialListingPage(ContactFieldsMixin, BasePage):
         for f in filters:
             queryset = f.apply(queryset, request.GET)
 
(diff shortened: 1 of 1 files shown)


rca/editorial/models.py
```diff
@@ -552,6 +552,8 @@ def get_context(self, request, *args, **kwargs):
         for f in filters:
             queryset = f.apply(queryset, request.GET)
 
+        queryset = queryset.distinct()
```
engineervix, line 555: good catch :+1:

engineervix (approved):
Thanks Patrick!

PR #1009, opened 2024-05-30
#3853 - Accessibility: Image 'ALT' text field - 'Decorative image' option

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1472503853

This PR adds a 'decorative' option to image blocks. I've checked other uses of images and it looks like we'll only need it for image blocks. All the rest will require alt text.


kevinhowbrook (approved):
:+1: just ci/black to fix then all good - thanks

PR #1010, opened 2024-05-31
#3766 - Accessibiity: Design System & Colour Contrast

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1472503766

This PR updates all xxxs, xxs, and xs font sizes to 18px. We also remove some `font-size: map-get($large-font-sizes, xxs);` as they're all now 18px.


chris-lawton:
The code looks good. I imagine that the components will need testing on staging to make sure that they work and display correctly with the increased font size : )

PR #1011, opened 2024-06-03
#6809 - Enable embed codes Intranet and Main Site

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1516186809
This MR adds an embed finder for Wix sites. Wix does not have oEmbed support so we implement a simple finder.


rca/utils/embed_finders.py
```diff
@@ -20,3 +24,27 @@ def find_embed(self, url, max_width=None):
...
+    def accept(self, url):
+        return re.match(r"^https?://(?:www\.)?embed.wix\.com/.+$", url)
+
+    def find_embed(self, url, max_width=None):
+        # Attempt to fetch the page content and scrape the title.
+        response = requests.get(url)
+        soup = BeautifulSoup(response.content, "html.parser")
+
+        title_tag = soup.find("title")
+        title = title_tag.string if title_tag else "No title found"
```
kevinhowbrook, line 41: As per intranet pr:

> Is "No title found" the right thing to return here? Wouldn't none be better to avoid rendering it in the template?

patrickcuagan: Makes sense! I've made it `None` now. Might make it confusing too if the title of the embed is "No title found".

PR #1020, opened 2024-07-29
Fix broken layout on short course pages

siimonevans:
The text within the booking bar often sits on one line. For example: https://www.rca.ac.uk/study/programme-finder/service-design-masterclass/

However, when the label is 'Register your interest for upcoming dates', it flows onto two lines. The page layout breaks when this happens, as heights/margins/etc assume the booking bar is shallower. For example, see the whitespace below the hero here: https://www.rca.ac.uk/study/programme-finder/human-centred-design-for-intelligent-mobility/

This PR aims to fix that. Creating a new modifier seemed like the easiest way, and we want to fix this asap since it's an issue on prod.

### DIFF
--- a/rca/static_src/sass/components/_page.scss
+++ b/rca/static_src/sass/components/_page.scss
@@ -73,6 +73,14 @@
                 margin-top: -91px;
             }
         }
+
+        &--with-sticky-cta-tall {
+            margin-top: -75px;
+
+            @include media-query(medium) {
+                margin-top: -117px;
+            }
+        }
     }
 
     &__notch-block {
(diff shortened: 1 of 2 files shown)


rca/static_src/sass/components/_page.scss
```diff
@@ -120,23 +128,6 @@
         }
     }
 
-    .sticky-bar.app--short-course & {
```
siimonevans, line 123: As far as I can tell, this doesn't do anything. Whitespace is already created via padding on the block wrappers, and with this in place text is obscured on mobile. E.g [on this page](https://www.rca.ac.uk/study/programme-finder/service-design-masterclass/), the text `Use the principles and tools of service design to rethink the journey your customers take and transform your customer experience, delivered live online.` is hidden.

Removing it didn't appear to harm anything, but it'd be good to double check with someone who has more context of the project.

nicklee: My understanding was that the sticky book bar used to sit at the top of the page, so when the mini header came in, it had to move down.

It appears to have since been fixed to the bottom, so this is no longer required.

nicklee:
LGTM! It's assuming that anything that isn't Book now is going to wrap over more than one line, so we just need to be careful that this is always the case?

siimonevans:
@nicklee great catch, thanks. Looks like there's a third condition for 'Apply now' (application forms) which I hadn't spotted previously. 

I've swapped the logic so that we target the longer string, meaning every other scenario else gets the same treatment as before 👍

PR #1021, opened 2024-08-02
#7446 : Add Table block to guide pages

ababic:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1538457446

Som quirks I've noticed in Wagtail, which aren't really in-scope for this MR:
- Wagtail seems not to automatically re-render previews as expected when the 'header' option checkboxes change
- The block appears not to have a bottom margin, so gets quite close to the checkbox below once the table is populated


kevinhowbrook (approved):

kevinhowbrook (approved):
Thanks @ababic - lgtm, only thing I would be tempted todo is squash the 2 migrations

PR #998, opened 2024-04-24
Python 3.11

katdom13:

PR #1030, opened 2024-08-06
Upgrade python to 3.11

nickmoreton:
Replace PR: https://github.com/torchbox/rca-wagtail-2019/pull/998

### When merged this will upgrade python to 3.11 using the bookworm image tag

Additions.

- Upgrade poetry
- Upgrade dependencies reported by trivy
- Add pyupgrade + pre-commit hook
- Add poetry pre-commit hook

Changes:
- Update upgrading docs


docs/upgrading.md
nickmoreton, line 1: I've done some clean up on this file. We don't need to keep the notes about previous Wagtail versions.

engineervix:
The python version can also be updated here:

https://github.com/torchbox/rca-wagtail-2019/blob/c8253243ad47627285f08141a0d3005ebaafda72/README.md?plain=1#L61

.python-version
```diff
@@ -1 +1 @@
-3.8.6
\ No newline at end of file
+3.11
```
engineervix, line 1: :ok_hand: it's nice that we are not locking down to a specific patch version. This will capture whatever 3.11.x version is availble.

engineervix (commented):
Thanks @nickmoreton changes look good to me and running ok locally.
Just a small update in the README to specify the required python version.

PR #1038, opened 2024-08-08
[7159] Make meta title and description fields required (and simplify templates)

ababic:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1588777159
Related ticket: https://torchbox.monday.com/boards/1472452416/pulses/1589345904

### DIFF
--- a/rca/project_styleguide/templates/patterns/base_page.html
+++ b/rca/project_styleguide/templates/patterns/base_page.html
@@ -2,6 +2,9 @@
 
 {% load static wagtailuserbar wagtailcore_tags wagtailimages_tags navigation_tags util_tags %}
 
+{% block title %}{{ page.meta_title }}{% endblock %}
+{% block meta_description %}{{ page.meta_description }}{% endblock %}
+
 {% block meta_tags %}
     {% if GOOGLE_TAG_MANAGER_ID %}
         {# To enable GTM code you need to specify GOOGLE_TAG_MANAGER_ID in Heroku settings #}
--- a/rca/project_styleguide/templates/patterns/base_page.html
+++ b/rca/project_styleguide/templates/patterns/base_page.html
@@ -6,26 +6,6 @@
 {% block meta_description %}{{ page.meta_description }}{% endblock %}
 
 {% block meta_tags %}
-    {% if GOOGLE_TAG_MANAGER_ID %}
-        {# To enable GTM code you need to specify GOOGLE_TAG_MANAGER_ID in Heroku settings #}
-        <script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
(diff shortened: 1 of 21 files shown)


rca/project_styleguide/templates/patterns/base_page.html
```diff
@@ -2,27 +2,10 @@
...
-    {# favicons #}
-    <link rel="apple-touch-icon" sizes="180x180" href="{% static "images/apple-touch-icon.png" %}">
-    <link rel="icon" type="image/png" sizes="32x32" href="{% static "images/favicon-32x32.png" %}">
-    <link rel="icon" type="image/png" sizes="16x16" href="{% static "images/favicon-16x16.png" %}">
-    <link rel="manifest" href="{% static "images/site.webmanifest" %}">
-    <link rel="mask-icon" href="{% static "images/safari-pinned-tab.svg" %}" color="#000000">
-    <link rel="shortcut icon" href="{% static "images/favicon.ico" %}">
-    <meta name="msapplication-TileColor" content="#000000">
-    <meta name="msapplication-config" content="{% static "images/browserconfig.xml" %}">
-    <meta name="theme-color" content="#000000">
```
kevinhowbrook, line 24: **food for though/ramblings**

I guess this is what prompted your recent comment about better use of base and base_page templates. I think the caveat is that we don't usually rely on base.html ro render elements with no `page` context and that's why we see it so ofter - I've never understood it tbh but I think your change does make sense, but I don't know if favicons belong in `base.html`, would we ever want to render them on something that's _not_ a page? Again it's probably moot because we don't use templates that way

rca/utils/forms.py
```diff
@@ -0,0 +1,26 @@
+from django.utils.html import mark_safe
+from wagtail.admin.forms import WagtailAdminPageForm
+
+
+class RCAPageAdminForm(WagtailAdminPageForm):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if seo_title_field := self.fields.get("seo_title"):
+            seo_title_field.required = True
```
kevinhowbrook, line 10: **Praise**

Nice way of doing this, TIL

kevinhowbrook (approved):
Thanks Andy, nice clean-ups :+1:

rca/project_styleguide/templates/patterns/base_page.html
ababic: > I guess this is what prompted your recent comment about better use of base and base_page templates.

Yes, but it did bug me on a couple of projects prior to this - it's just something I didn't really have time to address before :)

>  but I don't know if favicons belong in base.html, would we ever want to render them on something that's not a page? Again it's probably moot because we don't use templates that way

I would say so, yes. It's important for all HTML responses to retain the original branding - which favicons are a part of.

PR #1039, opened 2024-08-13
[0455] Allow alumni articles on homepage to be curated by editors

ababic:
Ticket: https://torchbox.monday.com/boards/1472452416/views/14036164/pulses/1589580455


kevinhowbrook (approved):
Looks good, thanks Andy. I'd plea for a max_num limit on the orderable but not imperative.

PR #1048, opened 2024-09-19
#6447 - CTA link style

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1509716447

This PR adds a "CTA Link" block to be used in guide, editorial, and event pages. They'll just follow the same CTA style within the build.

Screenshots:

Screenshot 2024-09-18 at 9 42 56 PM
Screenshot 2024-09-18 at 9 43 06 PM


rca/project_styleguide/templates/patterns/molecules/streamfield/blocks/cta_link_block.html
```diff
@@ -0,0 +1,17 @@
...
+        <svg class="link__icon" width="12" height="8" aria-hidden="true"><use xlink:href="#arrow"></use></svg>
+    </a>
+{% else %}
+    <a class="link link--primary link--link link--external link--block" href="{{ value.url }}" target="_blank">
+        <span class="link__label link__label--tight">
+            {{ value.title }}
+        </span>
+        <svg class="link__icon" width="12" height="8" aria-hidden="true"><use xlink:href="#arrow-external"></use></svg>
+    </a>
+{% endif %}
```
patrickcuagan, line 17: Just repeated the template to avoid having too much `if` statements within the html.

rca/static_src/sass/components/_link.scss
```diff
@@ -242,6 +242,19 @@
         max-width: 170px;
     }
 
+    &--block {
+        color: $color--tertiary !important;
```
siimonevans, line 246: Hey @patrickcuagan can we avoid the `!important`s?

I wonder if we need this to work against both light and dark backgrounds, too?

kevinhowbrook (approved):
Thanks @patrickcuagan, looks like just simons point to address her, BE looks good.

rca/static_src/sass/components/_link.scss
patrickcuagan: On guide, editorial, and event pages, the streamfield body has a light background so there's no need to support for dark backgrounds.

As for the `!important` -- this looks to be needed as the streamfield body sometimes (for guide and editorial) is enclosed with `rich-text`. 

```
<div class="streamfield rich-text layout__start-one layout__span-two layout__@large-start-two layout__@large-span-three">
                        {% for block in page.body %}
                            {% include_block block %}
                        {% endfor %}
                    </div>
```

Think removing the `.rich-text` from the parent div is risky. Not sure what other alternatives there are aside from adding the `important`.

siimonevans: @patrickcuagan ah right I see. If there's no way around it then perhaps we need to add an ignore for the linter at least?

patrickcuagan: Don't think we have a linter set up. There's quite a lot of !important throughout the build.

PR #1050, opened 2024-09-20
#7888 - BUG | Home 'Hero CTA sub text'

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1630887888

This PR fixes an issue where the hero CTA sub text is cut off on mobile.

Screenshot:

Screenshot 2024-09-20 at 3 06 50 PM

### DIFF
--- a/rca/static_src/sass/components/_hero-action-pane.scss
+++ b/rca/static_src/sass/components/_hero-action-pane.scss
@@ -1,6 +1,6 @@
 .hero-action-pane {
     $root: &;
-    $hero-action-pane-height-small: 110px;
+    $hero-action-pane-height-small: 190px;
     $hero-action-pane-height-medium: 160px;
     $hero-image-credit-height: 40px;
     $hero-action-pane-overlap: 80px;


rca/static_src/sass/components/_hero-action-pane.scss
```diff
@@ -1,6 +1,6 @@
 .hero-action-pane {
     $root: &;
-    $hero-action-pane-height-small: 110px;
+    $hero-action-pane-height-small: 190px;
```
patrickcuagan, line 3: I explored options where we don't hard code this but think we're going to have to refactor quite a bit if we wanted to.

siimonevans: Hey @patrickcuagan I think we should explore this further with RCA.

If we make this change and the content they add to these fields is later reduced, we could end up with the content appearing to high up:

Screenshot 2024-09-20 at 08 25 50

siimonevans: Agreed to proceed with this, but we'll create a separate ticket for refactoring this component in future.

PR #1051, opened 2024-09-20
#4453 - BUG | FE | Programme page whitespace

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1559054453

This PR fixes a white bar issue for short course pages.

Screenshot:

Screenshot 2024-09-20 at 5 59 19 PM

### DIFF
--- a/rca/static_src/sass/components/_booking-bar.scss
+++ b/rca/static_src/sass/components/_booking-bar.scss
@@ -111,4 +111,23 @@
             }
         }
     }
+
+    .app--short-course & {
+        &-last-item {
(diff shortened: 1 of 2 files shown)


rca/static_src/sass/components/_booking-bar.scss
```diff
@@ -111,4 +111,23 @@
             }
         }
     }
+
+    .app--short-course & {
```
patrickcuagan, line 115: It looks like we really have custom margin bottoms depending on the page based on the code above.

PR #1055, opened 2024-10-15
#3769 - Short course template > Add dates to key details

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1472503769

This PR adds a "Dates" field for short courses. This is then displayed in the key details area when viewing the page.

Screenshot 2024-10-15 at 1 57 43 PM


rca/account_management/templates/wagtailadmin/base.html
```diff
@@ -18,7 +18,7 @@
         A wagtail_hook adds some global JS and CSS to manage it's visibility in rca/people/wagtail_hooks.py
         {% endcomment %}
     </aside>
-    {% keyboard_shortcuts_dialog %}
+    {% comment %} {% keyboard_shortcuts_dialog %} {% endcomment %}
```
kevinhowbrook, line 21: Not sure what this is? Does it need to be here if it's commented out?

kevinhowbrook (approved):
Just a comment about the commented out template code, otherwise I think this is fine to merge on, thanks @patrickcuagan

rca/account_management/templates/wagtailadmin/base.html
patrickcuagan: It was raising an error when I access the admin so I commented it out.

This is needed though so I've added it back in.

PR #1068, opened 2025-01-16
#5479 - GUIDE, EDITORIAL, EVENT TEMPLATES | Add extra required fields

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1605295479

This MR makes the three fields required:

1. Social networks (image)
2. Social text
3. Listing information image

### DIFF
--- a/rca/utils/forms.py
+++ b/rca/utils/forms.py
@@ -24,3 +24,18 @@ class RCAPageAdminForm(WagtailAdminPageForm):
                 "content users can expect to find on the page. <strong>Recommended length: 50 to 160 "
                 "characters</strong>."
             )
+
+        # Listing Image field configuration
+        if listing_image_field := self.fields.get("listing_image"):
+            listing_image_field.required = True
+            listing_image_field.help_text = "This image is required for listings."
+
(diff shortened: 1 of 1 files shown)


rca/utils/models.py
```diff
@@ -256,7 +256,7 @@ class ListingFields(models.Model):
     listing_image = models.ForeignKey(
         "images.CustomImage",
         null=True,
-        blank=True,
+        blank=False,
```
patrickcuagan, line 259: Most of the files here are migrations. This is the only bit that's relevant.

kevinhowbrook (changes requested):
Thanks @patrickcuagan - Think the tests might need updating (CI failures) now given the required images

patrickcuagan:
> Thanks @patrickcuagan - Think the tests might need updating (CI failures) now given the required images

I've updated this to make it required via the form instead. I just remembered we briefly talked about this during refinement one time. This way - we don't have to update all the factories etc.

kevinhowbrook (approved):

PR #1069, opened 2025-01-17
#1301 - PROGRAMME FINDER | MRes Healthcare & Design and MRes RCA

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1605295479

This MR allows editors to add more than one programme_type for program and short course pages. This means that a program/short course can show in multiple program types when searching through the program finder.


rca/wagtailapi/views.py
```diff
@@ -20,6 +20,7 @@ class PagesAPIViewSet(views.PagesAPIViewSet):
         DescendantOfFilter,
         filters.RelatedSchoolsFilter,
         filters.SubjectsFilter,
+        filters.ProgrammeTypesFilter,
```
patrickcuagan, line 23: It's now a many to many field so we've had to add this additional filter.

kevinhowbrook (approved):
Thanks @patrickcuagan - This looks good to me.

I just want to point out, as I'm sure you are aware, that the enquire to study form has integrations with Mailchimp and QS... so when testing on dev (if possible still) we need to make sure the integration still accepts the data; although it doesn't look like the shape of data we send has changed. Just thinking if we send through program types, and that's now multiple it may need to be updated.

patrickcuagan:
> Thanks @patrickcuagan - This looks good to me.
> 
> I just want to point out, as I'm sure you are aware, that the enquire to study form has integrations with Mailchimp and QS... so when testing on dev (if possible still) we need to make sure the integration still accepts the data; although it doesn't look like the shape of data we send has changed. Just thinking if we send through program types, and that's now multiple it may need to be updated.

Yeah that was one of my concerns as well. For enquire to study forms, we use programme pages instead of the programme type. In programme pages, we we have a `qs_code` or `mailchimp_group_name` and that's what we pass into Mailchimp/QS.

PR #1070, opened 2025-01-22
Stop pulling news & events from the defunct legacy site

alxbridge:
As the legacy site is no longer online, calls to its content from the `refetch_legacy_news_and_events` fail, causing errors on pages whose templates render the legacy content. This fixes those errors by making `refetch_legacy_news_and_events` return an empty list instead, which will get things working until a longer-term code cleanup can take place.


kevinhowbrook (approved):
Thanks so much alex, that should do the trick

PR #1071, opened 2025-03-05
#7346 - WEB FORM | Register your interest - UK & IRE Data

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1851167346

This PR removes the Mailchimp integration that's no longer going to be used. Previously, data for users who entered UK/Ireland for their country of residences are sent to Mailchimp. All the rest goes to QS. This is now updated so that all data goes to QS no matter the country of residence chosen.

```
Currently, if the user selects either United Kingdom or Ireland as their Country of citizenship these countries are excluded.

So the requirement is to update this so that both countries are included in the data feed which is shared with QS.
```


rca/enquire_to_study/views.py
```diff
@@ -307,9 +205,8 @@ def set_session_data(self, form_data, enquiry_submission):
 
     def form_valid(self, form):
         country_of_residence = form.cleaned_data["country_of_residence"]
-        if country_of_residence in ["GB", "IE"] and settings.MAILCHIMP_API_KEY:
-            self.post_mailchimp(form.cleaned_data)
-        elif settings.QS_API_PASSWORD:
+
+        if settings.QS_API_PASSWORD:
```
patrickcuagan, line 209: This really is the only change. I noticed that Mailchimp won't be used anymore afterwards so I removed them all.

kevinhowbrook (approved):
Thanks @patrickcuagan, this looks fine to me.
And thank you for updating the docs

PR #1076, opened 2025-04-23
#3981 - EVENT PAGES | Enable Taxonomy for multiple types

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1811003981

This PR allows editors to add multiple event types to event pages.


rca/events/tests/test_api.py
```diff
@@ -160,7 +168,9 @@ def test_event_response(self):
             response.data["related_directorates"],
             [{"title": "ufos", "id": self.directorate.id, "intranet_slug": "u-f-o-s"}],
         )
-        self.assertEqual(response.data["event_type"]["title"], "party")
+        self.assertEqual(
+            response.data["event_types"], [{"id": self.event_type.id, "title": "party"}]
+        )
```
patrickcuagan, line 173: There's a change in the format - I don't see this being used in the site or in the intranet though.

kevinhowbrook: Thanks @patrickcuagan - had a look at why this was setup and there is an integration on the intranet for importing events https://git.torchbox.com/rca/intranet.rca.ac.uk/-/blob/master/inforca/admin/content_importer/static/content_importer/readme.md#L43

So it's worth double checking that importer still works ok with the new format

patrickcuagan: Thanks @kevinhowbrook - it looks to be fine. Event types aren't being imported to the intranet.

PR #1079, opened 2025-04-24
#3082 - IMAGES | Focal Point

patrickcuagan:
Ticket: https://torchbox.monday.com/boards/1472452416/pulses/1865283032

This PR updates the hero section so that the focal point is more evident.

Screenshots:

Screenshot 2025-04-24 at 1 39 33 PM

When focal point is:

Screenshot 2025-04-24 at 1 23 36 PM

### DIFF
--- a/rca/static_src/sass/components/_hero.scss
+++ b/rca/static_src/sass/components/_hero.scss
@@ -4,16 +4,16 @@
 
     &--medium {
         margin-top: 120px; // height of header
-        max-height: 220px; 
+        max-height: 170px; 
 
         @include media-query(medium) {
             margin-top: 130px;
-            max-height: 406px;
+            max-height: 320px;
         }
 
         @include media-query(large) {
             margin-top: 90px;
-            max-height: 690px;
+            max-height: 613px;
         }
     }
 
--- a/rca/static_src/sass/components/_hero.scss
+++ b/rca/static_src/sass/components/_hero.scss
@@ -3,14 +3,17 @@
     overflow: hidden;
 
     &--medium {
-        max-height: 340px;
+        margin-top: 120px; // height of header
+        max-height: 220px; 
 
         @include media-query(medium) {
-            max-height: 536px;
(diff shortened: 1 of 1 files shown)


rca/static_src/sass/components/_hero.scss
```diff
@@ -3,14 +3,17 @@
     overflow: hidden;
 
     &--medium {
-        max-height: 340px;
+        margin-top: 120px; // height of header
```
patrickcuagan, line 6: Seems like the only way I could solve this. The only side effect is that users will see a white bar when they scroll up when they're at the top most part of the page (which I don't think people do):

Production:

Screenshot 2025-04-24 at 1 41 33 PM

This change:

Screenshot 2025-04-24 at 1 41 36 PM

rca/project_styleguide/templates/patterns/molecules/hero/hero-schools.yaml
```diff
@@ -1,14 +0,0 @@
-tags:
```
patrickcuagan, line 1: Just noticed these unused files.

siimonevans (approved):

rca/static_src/sass/components/_hero.scss
patrickcuagan: As discussed on standup, we'll want to make the white bar to be black instead.

rca/static_src/sass/components/_hero.scss
```diff
@@ -3,14 +3,28 @@
...
+        &::before {
+            @include z-index(under);
+            content: '';
+            background-color: $color--black;
+            height: 130px;
+            left: 0;
+            position: absolute;
+            top: 0;
+            width: 100%;
+        }
```
patrickcuagan, line 15: Scrolled down on both desktop and mobile:

Screenshot 2025-04-24 at 7 05 33 PM

Screenshot 2025-04-24 at 7 05 40 PM

PR #1082, opened 2025-05-12
Fix search listings for promoted items with external links

alxbridge:
Fix for [incident reported 2025-05-12, 15:27](https://torchbox.monday.com/boards/1124499850/pulses/1954183790)

The template for _search.html_ passes promoted search items to _search-result.html_, with the assumption that there will be a `page` item attached which can be used to output the item in the results listing. This isn't true in the case of promoted items using external links, causing a template error, so this PR adds code to handle external promoted items.


nickmoreton (approved):
Great job @alxbridge looks good and works locally for me.

PR #1084, opened 2025-07-15
R1-229: BE - Flexible Home Page

patrickcuagan:
This pull request adds a `body` streamfield which will replace all the other fields in the home page. We kept the current home page fields and marked them as legacy to make deploying this piece of work safer (and easier).

Tickets:

- [Backend ticket](https://torchbox.atlassian.net/browse/R1-229)
- [Frontend ticket](https://torchbox.atlassian.net/browse/R1-230)

Screenshots:

Editor where you can select the background color + the available content blocks.

Screenshot 2025-07-15 at 4 01 42 PM


rca/home/models.py
```diff
@@ -152,6 +160,12 @@ class HomePage(TapMixin, BasePage):
     hero_cta_text = models.CharField(max_length=125, blank=True)
     hero_cta_sub_text = models.CharField(max_length=125, blank=True)
 
+    body = StreamField(
+        HomePageBodyBlock(),
+        blank=True,
+    )
+
+    # --- Legacy Body Fields --- #
```
kevinhowbrook, line 168: **Suggestion**

Remove this comment? or are we keeping the legacy fields? Perhaps worth updating the comment to say why we are keeping the legacy fields in?

rca/home/blocks.py
```diff
@@ -0,0 +1,208 @@
...
+            value["featured_event"]
+            and value["featured_event"].end_date > timezone.now().date()
+        ):
+            event = value["featured_event"]
+        else:
+            # If no featured event is selected or the selected event has ended,
+            # get the next upcoming event.
+            event = EventDetailPage.objects.filter(
+                end_date__gte=timezone.now().date()
+            ).order_by("start_date")[:1]
```
kevinhowbrook, line 153: **Suggestion**

Could this fail it no events exist?

Perhaps this might be more robust:

```
try:
    event = EventDetailPage.objects.filter(
        end_date__gte=timezone.now().date()
    ).order_by("start_date").first()
except Exception:
    event = None
```

rca/home/blocks.py
```diff
@@ -0,0 +1,208 @@
...
+
+        # If there is an event, we'll show 2 news items and 1 event.
+        NEWS_ITEMS = 2 if event else 3
+
+        news = (
+            EditorialPage.objects.filter(editorial_types__type__slug="news")
+            .live()
+            .filter(show_on_home_page=True)
```
kevinhowbrook, line 161: **Suggestion**

Worth adding  ` .select_related('hero_image')` here?

kevinhowbrook (commented):
Looking good @patrickcuagan - few suggestions...

Not sure how we are handling legacy fields? Are we removing them in another PR or leaving them in for a specific reason? I'm guessing it's so a new homepage can be populated but then later we remove the old fields once published?

patrickcuagan:
> Looking good @patrickcuagan - few suggestions...
> 
> Not sure how we are handling legacy fields? Are we removing them in another PR or leaving them in for a specific reason? I'm guessing it's so a new homepage can be populated but then later we remove the old fields once published?

Yep we'll be removing the legacy fields in this ticket: https://torchbox.atlassian.net/browse/R1-239.

rca/home/blocks.py
patrickcuagan: I've added this one - thank you!

PR #1095, opened 2025-07-30
Add trigram similarity search to search view

neilbulloch:
Closes https://torchbox.atlassian.net/browse/R1-245

`pg_trgm` is enabled via migrations in the search app, so it should be available for all deployment envs. 

The threshold currently is set at 0.1 (as a constant in rca/search/views.py, which is quite forgiving.

### DIFF
--- a/rca/search/views.py
+++ b/rca/search/views.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.contrib.postgres.search import TrigramSimilarity
 from django.core.paginator import EmptyPage, PageNotAnInteger, Paginator
 from django.template.response import TemplateResponse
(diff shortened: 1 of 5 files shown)


rca/search/views.py
```diff
@@ -19,6 +23,17 @@ def search(request):
 
         # Record hit
         query.add_hit()
+
+        # If no results found, try trigram similarity search
```
patrickcuagan, line 27: **Suggestion**

Can we add a small comment that we can only filter across `title` when using trigram similarity since that's the only common field we have among all page types?

patrickcuagan (approved):

PR #1098, opened 2025-07-31
Add programme location url field

neilbulloch:
Closes https://torchbox.atlassian.net/browse/R1-249

Added URL field in ProgramLocationProgramPage


rca/programmes/models.py
```diff
@@ -932,8 +933,8 @@ def study_mode(self):
 
     @cached_property
     def campus_locations(self):
-        return self.programme_locations.values_list(
-            "programme_location__title", flat=True
+        return self.programme_locations.values(
```
patrickcuagan, line 936: **Suggestion**

Wonder if it's worth doing a `prefetch_related("programme_location")` here to make it a little efficient.

PR #1139, opened 2026-02-13
Tidy up: use staging branches for staging deployments

kevinhowbrook:
The aim here is to make this set up less confusing... some tasks as part of retiring the dev site were not finished.

Dev was previously deploying to staging site - this PR is the final part of a tidy up to remove dev and have staging branch deploy to staging.

Legacy Intranet docs have also been updated too.


JakubMastalerz (approved):
Looks good! A couple of small notes.

PR #1140, opened 2026-02-24
Add GeoIP workaround to avoid exceptions when database is missing

kevinhowbrook:
## Problem

  Pages using `OriginCountryRule` personalisation segments were throwing 500 errors:

  django.contrib.gis.geoip2.GeoIP2Exception: GeoIP path must be provided
  via parameter or the GEOIP_PATH setting.

  The call chain is:

  1. `OriginCountryRule.get_country()` first tries `get_cloudflare_country()`, which reads
   the `CF-IPCountry` header — this returns `None` on certain requests (e.g. admin
  preview) where the header is absent
  2. It falls through to `get_geoip_country()`, which calls `get_geoip_module()` to import
   Django's `GeoIP2` class
  3. `get_geoip_module()` only catches `ImportError` — but Django GIS *is* installed on
  Heroku, so the import succeeds
  4. `GeoIP2()` is then instantiated, which raises `GeoIP2Exception` because `GEOIP_PATH`
  is not configured
  5. This exception is uncaught, producing a 500

  ## What we tried first (and why it didn't work)

  We added a `WAGTAIL_PERSONALISATION_IP_FUNCTION` setting pointing to a custom
  `rca/utils/geoip.py` that read the real client IP from Cloudflare's `CF-Connecting-IP`
  header.

  This setting controls the IP *extraction* step used inside `get_geoip_country()` — but
  `GeoIP2()` blows up during instantiation, *before* the IP function is ever called. So
  the custom function had no effect on the error.

  ## What we've done instead

  In `PersonalisationConfig.ready()`, we monkey-patch
  `OriginCountryRule.get_geoip_country` to wrap the original method in a `try/except`. If
  `GeoIP2Exception` is raised, it returns `None` instead of propagating the error —
  causing the country rule to simply not match rather than producing a 500.

  We check the exception by class name rather than importing `GeoIP2Exception` directly,
  because its location within `django.contrib.gis.geoip2` differs between Django versions
  (a single module file in some, a package in others).

  The `WAGTAIL_PERSONALISATION_IP_FUNCTION` setting and `geoip.py` have been removed as
  they were dead code.


rca/personalisation/apps.py
```diff
@@ -3,3 +3,37 @@
...
+        # Django GIS *is* installed, so get_geoip_module() returns the GeoIP2
+        # class successfully.  GeoIP2() then raises GeoIP2Exception at __init__
+        # because GEOIP_PATH is not configured, and that exception propagates
+        # uncaught, producing a 500.
+        #
+        # The Cloudflare header (HTTP_CF_IPCOUNTRY) is the preferred lookup path
+        # and normally short-circuits before GeoIP is reached.  This patch adds a
+        # safety net for requests where that header is absent (e.g. admin preview,
+        # local dev): get_geoip_country returns None, causing the country rule to
+        # not match rather than raising an error.
```
kevinhowbrook, line 22: Returning `none` before the exception is the fix here

nickmoreton (approved):

PR #1142, opened 2026-04-01
[R1-297] SCHOLARSHIP SNIPPETS | Character Limit + Links

katdom13:
JIRA Ticket: https://torchbox.atlassian.net/browse/R1-297
This PR:
- Makes the Scholarship summary field into rich text
- Migrates the Scholarship plain text summary values into the new rich text field, formatting the URLs in the process.
- Allows the Taxonomy titles and slugs to be longer

Screenshots

| Page | Before | After |
|---|---|---|
| https://www.rca.ac.uk/study/apply-to-study/funding-your-studies/rca-scholarships-and-awards/?programme=mres-rca#results |image | image |
| https://www.rca.ac.uk/study/programme-finder/mres-rca/#fees-funding | image | image |


rca/scholarships/models.py
```diff
@@ -67,9 +67,12 @@ class Meta:
 
 @register_snippet
 class Scholarship(models.Model):
-    title = models.CharField(max_length=50)
+    title = models.CharField(max_length=255)
```
katdom13, line 70: The ticket states:
> remove character limits from Scholarship title fields and other criteria taxonomy.

But we can't really do that, so the simplest solution would just be to make the `max_length` bigger.

PR #1144, opened 2026-04-07
R1-321: ScholarshipListingPage - Add FeeStatusFilter class for Home/Overseas filtering

patrickcuagan:
Ticket: https://torchbox.atlassian.net/browse/R1-321

This PR adds a `FeeStatusFilter` to replace the default inline `TabStyleFilter`. For this ticket, UI-wise, it doesn't look different.

### DIFF
--- a/rca/scholarships/filters.py
+++ b/rca/scholarships/filters.py
@@ -11,11 +11,19 @@ class FeeStatusFilter:
             False  # no tab/takeover panel — toggle switch handles this filter
         )
 
+    # The query parameter values are `uk` and `international`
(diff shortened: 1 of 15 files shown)


rca/scholarships/filters.py
```diff
@@ -1,6 +1,53 @@
 from rca.utils.filter import TabStyleFilter
 
 
+class FeeStatusFilter:
```
kevinhowbrook, line 4: I would have thought this should inherit `TabStyleFilter`?

kevinhowbrook (approved):

PR #1154, opened 2026-05-05
R1-334: Can't select certain pages to add personalisation CTAs to

patrickcuagan:
Ticket: https://torchbox.atlassian.net/browse/R1-334

This PR allows editors to display personalised CTAs in non-page views e.g. the enquiry to study form/thank you page. We only have that custom view so we just hard code it in the list of available page types as well.


JakubMastalerz (approved):
LGTM code-wise, and it's great that docs were updated. There are no clear ACs on the ticket, so I can't really say if this should be tested additionally?

PR #1155, opened 2026-05-07
R1-336: bump wgatail and django security upgrades

neilbulloch:
Bump django to 5.2.14 and wagtail to 7.3.2, fix CVE vulnerabilities


damwaingames (approved):
LGTM

